PR #1, opened 2019-01-04
Remove API definition and add Endpoint service

kenoir:
You cannot share custom domains between accounts (nor duplicate a custom domain in a 2nd account) - consequently in order to separate the application resources from the platform account - the rest API must remained defined in the same account as the custom domain (in this case the platform account).


kenoir:
Closing as _won't work!_

PR #8, opened 2019-01-15
Move bagger buckets to storage account for simplicity

kenoir:


kenoir:
@tomcrane they do - this is because the resource names in terraform conventionally use underscores - but the actual bucket names _are_ using hyphens.

PR #10, opened 2019-01-16
Scale down ECS

mklander:
Reduce the number of ECS instances and tasks running


kenoir:
Part of: https://github.com/wellcometrust/platform/issues/3272

PR #30, opened 2019-02-13
Fix more capitalisation problems in the bagger

alexwlchan:
For https://github.com/wellcometrust/platform/issues/3420

### DIFF
--- a/bagger/src/files.py
+++ b/bagger/src/files.py
@@ -29,11 +29,13 @@ def download_s3_object(b_number, source_bucket, source, destination):
         #
         # We use 'set()' to avoid trying the same key multiple times.
         #
-        alternative_bnums = set([
-            b_number.replace("x", "X"),
-            b_number.replace("b", "B"),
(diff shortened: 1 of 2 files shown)


bagger/src/files.py
```diff
@@ -15,10 +17,47 @@
...
+        # and the ALTO reference.  In that case, we should try some alternative
+        # keys (by fiddling with the case) to see if one of the alternatives works.
+        #
+        # We use 'set()' to avoid trying the same key multiple times.
+        #
+        alternative_bnums = set([
+            b_number.replace("x", "X"),
+            b_number.replace("b", "B"),
+            b_number.replace("x", "X").replace("b", "B"),
+        ]) - set([b_number])
```
mklander: Very small thing, I think if x is upper case b is also upper case, so you could increase the chances of a hit by:
```
[
            b_number.replace("x", "X"),
            b_number.replace("x", "X").replace("b", "B"),
            b_number.replace("b", "B"),
]
```

PR #38, opened 2019-02-15
Modify the Notifier to be sans-Akka

alexwlchan:
This patch rewrites the notifier to run sans-Akka streams/flows, and be a traditional WorkerService-backed app like the rest of the pipeline.

The code is simpler, nearly 100 lines shorter, and features several new tests to boot!


kenoir:
Nice basis for continuing work - before modifying existing apps I think we should talk through the design of `WorkerService` here. It makes sense to reuse an existing pattern for now but the runnable is a base for all apps and speaks to our future intentions for the whole platform.

alexwlchan:
> It makes sense to reuse an existing pattern for now but the runnable is a base for all apps and speaks to our future intentions for the whole platform.

Yep, makes sense.

We should definitely think about whether WorkerService is the correct pattern, long-term – but I’m also aware that we have a lot of stuff in the air already, which is why I copied it straight over without tweaking it too much for this code. In the short-term I want us to have less ways to write apps, not more. :)

PR #44, opened 2019-02-20
Rename progress/registrar namespaces to ingests/bags

kenoir:
🌲 
⬇️ 
🍓


alexwlchan:
Also, worth renaming into the "storage" namespace while we're here?

PR #45, opened 2019-02-20
Copy the entire contents of a bag in the replicator, not just the first 1000 objects

alexwlchan:
I found a paginator in the V1 API, and so I wrote some super-simple code that blocks when copying every single object. Includes a test that it works for lots of objects, and that if an individual object fails, we fail the entire copy.

I’ve also added a new `ObjectCopier` trait, because making this abstract opens the door to copying from S3 to elsewhere later (e.g. Azure).

Resolves https://github.com/wellcometrust/platform/issues/3450

### DIFF
--- a/bag_replicator/src/main/scala/uk/ac/wellcome/platform/archive/bagreplicator/storage/S3PrefixCopier.scala
+++ b/bag_replicator/src/main/scala/uk/ac/wellcome/platform/archive/bagreplicator/storage/S3PrefixCopier.scala
@@ -3,21 +3,14 @@ package uk.ac.wellcome.platform.archive.bagreplicator.storage
 import com.amazonaws.services.s3.AmazonS3
 import com.amazonaws.services.s3.iterable.S3Objects
 import com.amazonaws.services.s3.model.S3ObjectSummary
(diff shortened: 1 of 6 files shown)


bag_replicator/src/main/scala/uk/ac/wellcome/platform/archive/bagreplicator/storage/S3PrefixCopier.scala
```diff
@@ -28,66 +28,28 @@ class S3PrefixCopier(s3Client: AmazonS3)(implicit ec: ExecutionContext)
...
+    // enough in practice.
 
-  private def listObjectsUnderPrefix(
-    locationPrefix: ObjectLocation
-  ): Future[List[ObjectLocation]] = {
-    // TODO: limit size of the returned List
-    // use Marker to paginate(?)
-    // needs care if bag contents can change during copy.
-    debug(s"listing items in $locationPrefix")
+    while (objects.hasNext) {
```
kenoir: Can you just do `.map` here?

alexwlchan: oh yeah 😂

PR #48, opened 2019-02-21
Support alternative archive format for bags in bagger

kenoir:
In order to use a better suited container and compression format for the storage service (tar/gz) the bagger needs to provide bags in that format.

This updated the bagger to support by configuration any format supported here: https://docs.python.org/3/library/shutil.html#shutil.make_archive

This configuration is set by environment variable `ARCHIVE_FORMAT` and defaults to `zip`.


bagger/src/bag_assembly.py
```diff
@@ -67,4 +62,5 @@ def get_separator():
 def ensure_directory(destination):
     destination_dir = os.path.dirname(destination)
     if not os.path.exists(destination_dir):
-        os.makedirs(destination_dir)  # TODO: beware race condition if threaded
+        os.makedirs(destination_dir)
+        # TODO: beware race condition if threaded
```
alexwlchan: minor: not sure why we're fiddling with this code, but if you are, you might change this to

```python
os.makedirs(destination_dir, exist_ok=True)
```

which is race-free and simpler!

bagger/src/bagger.py
```diff
@@ -147,23 +153,37 @@ def load_xml(path):
     return load_from_string(xml_string)
 
 
-def dispatch_bag(bag_details):
-    # now zip this bag in a way that will be efficient for the archiver
-    logging.debug("creating zip file for " + bag_details["b_number"])
+def pack(bag_details):
+    archive_format = settings.ARCHIVE_FORMAT
+
+    logging.debug(" ".join(["packing", archive_format, bag_details["b_number"]]))
```
alexwlchan: nooooooo

I find this approach to logging super unclear, it's also less efficient in Python (because we construct the string even if we don't log it). I prefer:

```python
logging.debug("packing %s %s", archive_format, bag_details["b_number"])
```

bagger/src/bagger.py
```diff
@@ -147,23 +153,37 @@ def load_xml(path):
...
 
-    shutil.make_archive(
-        bag_details["zip_file_path"][0:-4], "zip", bag_details["directory"]
+    base_name = "/".join([settings.WORKING_DIRECTORY, bag_details["b_number"]])
```
alexwlchan: minor: when you're constructing paths, I think it's neater to use
```suggestion
    base_name = os.path.join(settings.WORKING_DIRECTORY, bag_details["b_number"])
```

bagger/src/bagger.py
```diff
@@ -147,23 +153,37 @@ def load_xml(path):
...
+
+    zip_file_path = shutil.make_archive(
+        base_name, archive_format, bag_details["directory"]
```
alexwlchan: Keyword arguments so it's clear what this is doing please?

bagger/src/bagger.py
```diff
@@ -147,23 +153,37 @@ def load_xml(path):
...
     )
 
-    logging.debug("uploading " + bag_details["zip_file_name"] + " to S3")
+    zip_file_name = zip_file_path.split("/")[-1]
```
alexwlchan: nicer:
```suggestion
    zip_file_name = os.path.basename(zip_file_path)
```

bagger/src/bagger.py
```diff
@@ -147,23 +153,37 @@ def load_xml(path):
...
+
 
+def dispatch(bag_details):
     upload_location = aws.upload(
         bag_details["zip_file_path"], bag_details["zip_file_name"]
     )
 
-    logging.debug("upload completed")
+    logging.debug(
+        " ".join(["uploaded", bag_details["zip_file_name"], f"{upload_location}"])
```
alexwlchan: ditto above, I'd prefer
```suggestion
        "uploaded %s %s", bag_details["zip_file_name"], upload_location)
```

also, `f"{upload_location}"` seems redundant?

bagger/src/bagger.py
kenoir: 👍

PR #55, opened 2019-02-21
BagRequest

kenoir:
Rename `ReplicationRequest` to `BagRequest` which will be how that operate on bags will get messages.


kenoir:
It's intentionally generic! I want the storage services where possible to communicate in the simplest format they can. The param names are weird as this is just `ReplicationRequest` renamed to `BagRequest`. 

I'm trying to push things gently in the direction of the split up archivist so when we create our new services they won't have to use names from the old world.

PR #66, opened 2019-02-25
Simplify StorageManifest

kenoir:
Simplifies the StorageManifest so that it's creation can be split into download/parse steps and the creation of component parts separated.


archivist/src/test/scala/uk/ac/wellcome/platform/archive/archivist/bag/ZippedBagFileTest.scala
```diff
@@ -39,29 +39,6 @@ class ZippedBagFileTest
       }
     }
 
-    it("fails if there is no valid bag-info file in zip file") {
```
kenoir: This test fails in `master` and i've no idea why - so i'm blatting it as I don't feel like it's worth fixing.

alexwlchan:
The Travis site reports this green, so merging.

PR #72, opened 2019-02-26
ConvertibleToInputStream returns Future

kenoir:
As we prefer this over a `Try`.


kenoir:
Closes https://github.com/wellcometrust/storage-service/pull/71

PR #88, opened 2019-03-01
Don't extend from Logging when we don't use it

alexwlchan:


kenoir:
Not touching the archivist!

PR #90, opened 2019-03-04
Log time taken in ISO-8601 format

mklander:
tiny change to output time taken by verification in ISO-8601, as this will show hours/seconds and subsecond part.


bag_verifier/src/main/scala/uk/ac/wellcome/platform/archive/bagverifier/services/BagVerifierWorkerService.scala
```diff
@@ -122,11 +122,10 @@ class BagVerifierWorkerService(
     }
     f"""$verificationStatus verification
        |of ${bagRequest.bagLocation.completePath}
-       |completed in ${bagVerification.duration.getSeconds}s
+       |completed in ${bagVerification.duration.getOrElse("")}
```
alexwlchan: I think this would be better we were really explicit about what's happening here. How about:

```suggestion
       |completed in ${bagVerification.duration.getOrElse("<NOTCOMPLETE>")}
```

bag_verifier/src/main/scala/uk/ac/wellcome/platform/archive/bagverifier/services/VerifyDigestFilesService.scala
```diff
@@ -49,15 +43,13 @@ class VerifyDigestFilesService(
     bagLocation: BagLocation,
     digestFiles: Seq[BagDigestFile]
   )(implicit materializer: Materializer): Future[BagVerification] = {
-    val verificationStart = Instant.now
-
+    val bagVerification = BagVerification(startTime = Instant.now)
```
alexwlchan: Perhaps we could make `Instant.now` then default value for this parameter?

kenoir: I think we discussed this and decided that it's better to be explicit when creating a verification what the start time is.

bag_verifier/src/main/scala/uk/ac/wellcome/platform/archive/bagverifier/services/BagVerifierWorkerService.scala
kenoir: :+1:

PR #94, opened 2019-03-05
Refactoring in the bag replicator

alexwlchan:
Part of https://github.com/wellcometrust/platform/issues/3443

I'm going to copy the existing bag_replicator wholesale, so tidying it up first felt useful. This patch should have no functional changes, just cleanups.


archivist/src/main/scala/uk/ac/wellcome/platform/archive/archivist/models/BagItemLocation.scala
```diff
@@ -1,7 +1,11 @@
-package uk.ac.wellcome.platform.archive.common.models.bagit
+package uk.ac.wellcome.platform.archive.archivist.models
```
kenoir: Nice. Quarantine those models!

PR #95, opened 2019-03-05
Add a bag replicator that goes from unpacked bags ~> archive bucket

alexwlchan:
Suggestions on naming welcome!

This patch adds a bag replicator that:

* Gets a BagRequest from SQS
* Finds the bag identifier and root of the bag in question
* Copies the bag into the archive bucket with the identifier as the key
* Sends a BagRequest with the new destination

Probably easiest if you clone locally and browse that way.


bag_replicator_unpack_to_archive/src/main/scala/uk/ac/wellcome/platform/archive/bagreplicator/unpack_to_archive/services/UnpackedBagService.scala
```diff
@@ -0,0 +1,31 @@
...
+import com.amazonaws.services.s3.AmazonS3
+import uk.ac.wellcome.platform.archive.common.ConvertibleToInputStream._
+import uk.ac.wellcome.platform.archive.common.bagit.{BagInfoLocator, S3BagFile}
+import uk.ac.wellcome.platform.archive.common.models.bagit.{BagInfo, ExternalIdentifier}
+import uk.ac.wellcome.platform.archive.common.parsers.BagInfoParser
+import uk.ac.wellcome.storage.ObjectLocation
+
+import scala.concurrent.{ExecutionContext, Future}
+
+class UnpackedBagService(s3Client: AmazonS3)(implicit ec: ExecutionContext) {
```
kenoir: I don't think this needs to share the idea of an unpacked bag. This is kind of a `BagLocator` maybe?

PR #102, opened 2019-03-06
Use hashed data structures to improve bagger performance

alexwlchan:
In Python, there's a big penalty for doing lookups in lists:

- `x in list(foo)` is O(n) in the length of the list
- `x in set(foo)` or `x in dict(foo)` is O(1)

By converting a couple of internal data structures to sets/dicts, we should make the bagger go a bit faster. How much faster? No idea, this is just picking some really low-hanging fruit.

(But for comparison, changing the data structures took the time to compute the destinations of ~400k Miro records from ~90 minutes to ~3 minutes, so it might help!)

### DIFF
--- a/bagger/src/bagger.py
+++ b/bagger/src/bagger.py
@@ -41,7 +41,7 @@ def bag_from_identifier(identifier, skip_file_download):
     root = tree.getroot()
     title = mets.get_title(root)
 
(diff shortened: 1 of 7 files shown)


kenoir:
@tomcrane can you take a look at this?

alexwlchan:
Replaced by #112.

PR #106, opened 2019-03-07
Fix directory handling in unpacker

kenoir:
The unpacker should ignore "./" at the beginning of entry paths and not attempt to create directories in S3.


kenoir:
@alexwlchan ideally more tests. Not now though.

PR #111, opened 2019-03-07
Fail the ingest if we're asked to unpack a non-existent bag; don't just fall over

alexwlchan:
Resolves https://github.com/wellcometrust/platform/issues/3479


bag_unpacker/src/main/scala/uk/ac/wellcome/platform/archive/bagunpacker/services/Unpacker.scala
```diff
@@ -50,6 +50,10 @@ class Unpacker(implicit s3Client: AmazonS3, ec: ExecutionContext) {
     } yield result
 
     futureSummary
+      .recover {
+        case err: Throwable =>
+          OperationFailure(unpackSummary.complete, err)
+      }
   }
```
mklander: 👍

PR #116, opened 2019-03-07
Use OperationNotifier in verifier

kenoir:


bag_verifier/src/main/scala/uk/ac/wellcome/platform/archive/bagverifier/Main.scala
```diff
@@ -31,37 +31,21 @@ object Main extends WellcomeTypesafeApp {
...
       storageManifestService = new StorageManifestService(),
       s3Client = s3Client,
       algorithm = MessageDigestAlgorithms.SHA_256
     )
 
-    val notificationService = new NotificationService(
-      progressSnsWriter,
-      outgoingSnsWriter
-    )
+    val operationName = "verification"
```
alexwlchan: nit: Should this be capitalised?

kenoir: Gets capitalised in the output :)

PR #117, opened 2019-03-08
Rename Progress -> Ingest

kenoir:
Give things name that better represent what they are.


bag_register/src/test/scala/uk/ac/wellcome/platform/archive/bag_register/BagRegisterFeatureTest.scala
```diff
@@ -69,10 +67,10 @@ class BagRegisterFeatureTest
 
             storageManifest.createdDate.isAfter(createdAfterDate) shouldBe true
 
-            assertTopicReceivesProgressStatusUpdate(
+            topicRecievesIngestStatus(
```
alexwlchan: typo: should be “Receives”

PR #118, opened 2019-03-08
Use a TransferManager to upload objects in the unpacker

alexwlchan:
A somewhat speculative fix for https://github.com/wellcometrust/platform/issues/3480.

You can’t upload more than 5GB in a single PutObject request to S3 – after that, you’re meant to switch to MultipartUpload. We weren’t doing that, hence why the really big uploads failed.

The Java SDK docs suggest we should be using TransferManager instead: https://docs.aws.amazon.com/AmazonS3/latest/dev/usingHLmpuJava.html

I’m struggling to test it because I can’t reproduce the error with our S3 Docker image (it seems to accept arbitrarily large file uploads with a single PutObject call, or at least bigger than I can create during a test run), but it doesn't break any existing tests. Worth a try?

### DIFF
--- a/bag_unpacker/src/main/scala/uk/ac/wellcome/platform/archive/bagunpacker/services/Unpacker.scala
+++ b/bag_unpacker/src/main/scala/uk/ac/wellcome/platform/archive/bagunpacker/services/Unpacker.scala
@@ -5,7 +5,7 @@ import java.nio.file.Paths
 import java.time.Instant
 
 import com.amazonaws.services.s3.AmazonS3
-import com.amazonaws.services.s3.model.{ObjectMetadata, PutObjectRequest}
+import com.amazonaws.services.s3.model.{ObjectMetadata, PutObjectRequest, PutObjectResult}
 import org.apache.commons.compress.archivers.ArchiveEntry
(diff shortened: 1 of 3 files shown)


alexwlchan:
If this works and we like it, I’d be tempted to push into into the storage-lib as an S3Uploader, and use it for all future PutObject operations.

kenoir:
I need to check how transfer manager deals with an InputStream here, to ensure it won't accumulate the stream in memory.

alexwlchan:
I checked this and think it’s okay, lemme find a reference.

alexwlchan:
https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/services/s3/transfer/TransferManager.html

> Schedules a new transfer to upload data to Amazon S3. This method is non-blocking and returns immediately (i.e. before the upload has finished).
>
> When uploading options from a stream, callers must supply the size of options in the stream through the content length field in the ObjectMetadata parameter. If no content length is specified for the input stream, then TransferManager will attempt to buffer all the stream contents in memory and upload the options as a traditional, single part upload. Because the entire stream contents must be buffered in memory, this can be very expensive, and should be avoided whenever possible.

PR #119, opened 2019-03-08
Give the bagger room to deploy new tasks

alexwlchan:
Because the bagger is running at the limit of its network interfaces, it can't deploy new task definitions – the old ones never go away, and it can't start new ones.

This frees up a single slot for the bagger to start a new task definition, so it can gradually update itself without needing us to go in and stop all the existing tasks.

### DIFF
--- a/terraform/stack/main.tf
+++ b/terraform/stack/main.tf
@@ -305,10 +305,10 @@ module "bagger" {
   cpu    = "1900"
   memory = "14000"
 
-  min_capacity = "${var.desired_bagger_count / 2}"
+  min_capacity = "${var.desired_bagger_count - 1}"
   max_capacity = "${var.desired_bagger_count}"
 
-  desired_task_count = "${var.desired_bagger_count}"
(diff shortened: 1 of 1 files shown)


kenoir:
You could also allow the bagger to scale down to 0% while deploying, so we don't waste resources.

terraform/stack/main.tf
```diff
@@ -305,10 +305,10 @@ module "bagger" {
   cpu    = "1900"
   memory = "14000"
 
-  min_capacity = "${var.desired_bagger_count}"
+  min_capacity = "${var.desired_bagger_count - 1}"
   max_capacity = "${var.desired_bagger_count}"
 
-  desired_task_count = "${var.desired_bagger_count}"
+  desired_task_count = "${var.desired_bagger_count - 1}"
```
mklander: `max_capacity = "${var.desired_bagger_count + 1}"`
looks more like the intention to me :)

PR #123, opened 2019-03-11
Operation notifier is breaking up the band

alexwlchan:


bag_register/src/main/scala/uk/ac/wellcome/platform/archive/bag_register/services/BagRegisterWorker.scala
```diff
@@ -6,15 +6,16 @@ import io.circe.Encoder
 import uk.ac.wellcome.json.JsonUtil._
 import uk.ac.wellcome.messaging.sqs.NotificationStream
 import uk.ac.wellcome.platform.archive.common.models.BagRequest
-import uk.ac.wellcome.platform.archive.common.operation.OperationNotifier
+import uk.ac.wellcome.platform.archive.common.operation.{OperationNotifier, OperationReporter}
```
kenoir: I'd rename these to remove the already namespaces "operation" - possibly call `Notifier`something that we don't already use somewhere else.

PR #128, opened 2019-03-12
Restructure to bring APIs together

kenoir:
I think the APIs belong together, moving towards a top level that contains:

- Services not specific to BagIt
- API namespace (containing display)
- A single "common" directory
- A BagIt namespace for services that operate on BagIt bags specifically.


alexwlchan:
Quick, merge it before it conflicts again!

PR #133, opened 2019-03-12
Outgoing response created in worker

mklander:
A suggestion to simplify the outgoing publisher - pass the required message in from the worker rather than passing a function to build it.


common/src/test/scala/uk/ac/wellcome/platform/archive/common/generators/BagRequestGenerators.scala
```diff
@@ -1,13 +1,23 @@
...
     BagRequest(
       requestId = randomUUID,
       bagLocation = bagLocation
     )
+
+  def createBagLocation() =
+    BagLocation(
+      storageNamespace = randomAlphanumeric(),
+      storagePrefix = Some(randomAlphanumeric()),
+      storageSpace = StorageSpace(randomAlphanumeric()),
```
alexwlchan: nit: I think there's a helper like `createStorageSpace`.

mklander: Thanks, so many Generators

common/src/main/scala/uk/ac/wellcome/platform/archive/common/operation/OutgoingPublisher.scala
```diff
@@ -1,42 +1,29 @@
...
-  )(implicit
+  def sendIfSuccessful[R, O](result: OperationResult[R], outgoing: => O)(
+    implicit
     ec: ExecutionContext,
-    enc: Encoder[O]): Future[Unit] = result match {
-    case OperationSuccess(summary) =>
-      sendOutgoing(transform(summary)).map(_ => ())
-    case OperationFailure(_, _) =>
+    enc: Encoder[O]): Future[Unit] = {
+    if (result.isSuccessful) {
```
kenoir: You can replace this with a match and `OR` the two success type together - that way you can avoid the case to boolean mapping below.

common/src/main/scala/uk/ac/wellcome/platform/archive/common/ingests/operation/OperationResult.scala
```diff
@@ -3,6 +3,16 @@ package uk.ac.wellcome.platform.archive.common.ingests.operation
 sealed trait OperationResult[T] {
   val summary: T
   def withSummary(summary: T): OperationResult[T]
+  def isSuccessful: Boolean = {
```
kenoir: This is _not_ very scala-esque.

PR #137, opened 2019-03-13
API via NLB task count must equal AZ count

kenoir:
In order to prevent performance issues.

### DIFF
--- a/terraform/stack/api/main.tf
+++ b/terraform/stack/api/main.tf
@@ -14,7 +14,8 @@ module "services" {
   // This is due to the behaviour of NLBs that
   // seem to increase latency significantly if
   // number of tasks < number of AZs
-  desired_bags_api_count    = "3"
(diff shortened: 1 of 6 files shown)


terraform/stack/api/main.tf
```diff
@@ -10,8 +10,12 @@ module "services" {
...
   nlb_arn    = "${module.nlb.arn}"
 
-  desired_bags_api_count    = "${var.desired_bags_api_count}"
-  desired_ingests_api_count = "${var.desired_ingests_api_count}"
+  // The number of api tasks MUST be one per AZ
+  // This is due to the behaviour of NLBs that
+  // seem to increase latency significantly if
+  // number of tasks < number of AZs
+  desired_bags_api_count    = "3"
+  desired_ingests_api_count = "3"
```
mklander: 😁

PR #149, opened 2019-03-19
S3Uploader set ReadLimit

mklander:
Set the S3 read limit to the size of the largest PUT so the stream can be reset.

https://github.com/wellcometrust/platform/issues/3481


bag_unpacker/src/main/scala/uk/ac/wellcome/platform/archive/bagunpacker/services/S3Uploader.scala
```diff
@@ -8,12 +8,31 @@ import com.amazonaws.services.s3.transfer.model.UploadResult
...
+  // occurs the mark will be invalid and a ResetException is thrown.
+  //
+  // When using a BufferedInputStream up to 'ReadLimit' bytes are stored in memory, and this must be (at least one byte)
+  // larger than the PUT.  The buffer grows dynamically up to this limit.
+  //
+  // For multipart PUT requests the size must be larger than each PART.
+  //
+  // To prevent this exception a constant maximum size is set
+  // assuming a maximum file of 600GB PUT as 10,000 multipart requests
+  // = 60MB ~ 100MB read limit
```
alexwlchan: I’d add a link to our ticket as well, so future readers can reconstruct our attempts to fix it.

And possibly a note that 100MB * 10 streams = 1GB, so we're comfortable this will fit into memory?

PR #160, opened 2019-03-28
Remove the distinction between access and archive locations

alexwlchan:
Closes wellcometrust/platform#3502.


alexwlchan:
> I'm not sure if this breaks what DDS/DLCS expect and hence whether we should merge it now though

It does; we should leave it until Tom and/or Adam are next in the office.

mklander:
We can merge this, Tom said it's a relatively small change and he will need to fix other things soon.

PR #161, opened 2019-04-02
Replicator: better logic for detecting the root of a bag

alexwlchan:
This simplifies the logic for detecting a bag. We now only detect a bag in one of two cases:

* The root of the archive is the bag
* There's exactly one directory in the archive, and it's the bag

Anything else is an error. This fixes #3517 and the issues Helen was having if the bag happens to contain a file called `bag-info.txt`.


bag_replicator/src/main/scala/uk/ac/wellcome/platform/archive/bagreplicator/services/S3BagLocator.scala
```diff
@@ -0,0 +1,107 @@
...
+  }
+
+  def locateBagRoot(objectLocation: ObjectLocation): Try[ObjectLocation] =
+    locateBagInfo(objectLocation).map {
+      loc => loc.copy(key = loc.key.stripSuffix("/bag-info.txt"))
+    }
+
+  /** Find a bag directly below a given ObjectLocation. */
+  private def findBagInfoInRoot(objectLocation: ObjectLocation): Try[String] = Try {
+    val listObjectsResult = s3Client.listObjectsV2(
```
mklander: Minor, but as we're only checking there isn't more than one we could limit the objects returned by setting `MaxKeys` in the request (to 2) -- but it's not so readable and shouldn't ever be a lot of keys.

`ListObjectsV2Request().withBucketName(objectLocation.namespace)withPrefix(createBagInfoPath(objectLocation.key).withMaxKeys(2)`

bag_replicator/src/main/scala/uk/ac/wellcome/platform/archive/bagreplicator/services/S3BagLocator.scala
```diff
@@ -0,0 +1,107 @@
...
+    }
+  }
+
+  def locateBagRoot(objectLocation: ObjectLocation): Try[ObjectLocation] =
+    locateBagInfo(objectLocation).map {
+      loc => loc.copy(key = loc.key.stripSuffix("/bag-info.txt"))
+    }
+
+  /** Find a bag directly below a given ObjectLocation. */
+  private def findBagInfoInRoot(objectLocation: ObjectLocation): Try[String] = Try {
```
mklander: `objectLocation` could be `bagLocation`

bag_replicator/src/main/scala/uk/ac/wellcome/platform/archive/bagreplicator/services/S3BagLocator.scala
alexwlchan: If you pass a delimiter you don’t get any keys back; `getObjectSummaries` is empty.

PR #164, opened 2019-04-03
Use the new Python client in our scripts; add a new script for looking up an ingest

alexwlchan:


bin/ingest_bag.py
```diff
@@ -33,136 +33,40 @@
...
-    print(description)
-    dumped_json = json.dumps(data, indent=2)
-
-    if data["status"]["id"] == "succeeded":
-        print(crayons.green(dumped_json))
-    elif data["status"]["id"] == "failed":
-        print(crayons.red(dumped_json))
-    else:
-        print(dumped_json)
+from wellcome_storage_service import StorageServiceClient
```
mklander: 😀

PR #165, opened 2019-04-03
Better error messages in progress updates

mklander:
user facing messages in ingest progress update from specific exceptions


bag_unpacker/src/main/scala/uk/ac/wellcome/platform/archive/bagunpacker/services/Unpacker.scala
```diff
@@ -36,35 +38,41 @@ case class Unpacker(s3Uploader: S3Uploader)(implicit s3Client: AmazonS3,
...
+                )
+                summary.copy(
+                  fileCount = summary.fileCount + 1,
+                  bytesUnpacked = summary.bytesUnpacked + archiveEntrySize
+                )
+              } catch {
+                case ae: AmazonS3Exception =>
+                  throw new PutObjectLocationException(
+                    dstLocation,
+                    "upload failed",
```
alexwlchan: This is the sort of exception I think we should be keeping private – how the storage service manages itself is an internal concern. If somebody sees “upload failed” in an unpacker message, how do they fix that?

common/src/main/scala/uk/ac/wellcome/platform/archive/common/ConvertibleToInputStream.scala
```diff
@@ -20,17 +21,21 @@ object ConvertibleToInputStream {
...
           response => response.getObjectContent
         )
         .recover {
           case throwable: Throwable =>
-            throw new RuntimeException(
-              s"Error getting input stream for s3://$t: ${throwable.getMessage}")
+            throw new InvalidObjectLocationException(
+              objectLocation = objectLocation,
+              message =
+                s"Error getting input stream for s3://$objectLocation: ${throwable.getMessage}",
```
alexwlchan: I think the more useful thing we could do here would be to distinguish between the most common S3 errors, and bubble up specific exceptions for those.

“No such object in S3” is one

“Do not have permissions to read object” is another

And this is where we need to make that distinction.

common/src/main/scala/uk/ac/wellcome/platform/archive/common/ingests/services/IngestUpdater.scala
```diff
@@ -62,4 +67,29 @@ class IngestUpdater(
...
+    val separator: String = " - "
+    truncate(
+      Seq(
+        Some(main),
+        maybeInformation
+      ).flatten.mkString(separator),
+      descriptionMaxLength)
+  }
+
+  private def truncate(text: String, maxLength: Int): String = {
```
alexwlchan: This is good, but I’d throw in a warning here if we're dropping descriptions that are >250 characters long, because we really shouldn't.

(This means not including S3 keys in messages, for example, although I think we're already doing that.)

bag_unpacker/src/main/scala/uk/ac/wellcome/platform/archive/bagunpacker/services/Unpacker.scala
mklander: Yup, looking again I think these exceptions should be specific to unpacking.

common/src/main/scala/uk/ac/wellcome/platform/archive/common/ConvertibleToInputStream.scala
mklander: 👍 distinguishing these by status code.

bag_register/src/test/scala/uk/ac/wellcome/platform/archive/bag_register/fixtures/BagRegisterFixtures.scala
```diff
@@ -53,8 +54,9 @@ trait WorkerFixture
                           withOutgoingPublisher("register", outgoingTopic) {
                             outgoingPublisher =>
                               val service = new BagRegisterWorker(
-                                alpakkaSQSWorkerConfig =
-                                  createAlpakkaSQSWorkerConfig(queuePair.queue),
+                                alpakkaSQSWorkerConfig = AlpakkaSQSWorkerConfig(
```
alexwlchan: nit: you can use `createAlpakkaSQSWorkerConfigWith(queuePair.queue)`

bag_unpacker/src/main/scala/uk/ac/wellcome/platform/archive/bagunpacker/services/BagUnpackerWorker.scala
```diff
@@ -40,15 +52,45 @@ case class BagUnpackerWorker(alpakkaSQSWorkerConfig: AlpakkaSQSWorkerConfig,
...
+        IngestFailed(s, e)
     }
+  }
+
+  private def clientMessageFor(exception: ArchiveLocationException) = {
+    val cause = exception.getCause.asInstanceOf[AmazonS3Exception]
+    val archiveLocation = exception.getObjectLocation
+    cause.getStatusCode match {
+      case 403 => s"access to $archiveLocation is denied"
+      case 400 => s"$archiveLocation is invalid"
```
alexwlchan: I might drop the 400 error code, because “invalid” isn’t a so helpful message, and there might be other S3 exceptions that drop this code? Others look fine.

PR #176, opened 2019-04-16
Don't do a toJson(…).get in the notifier

alexwlchan:
Spotted by accident; I think a JSON encoding error here would throw an exception.

### DIFF
--- a/notifier/src/main/scala/uk/ac/wellcome/platform/archive/notifier/services/CallbackUrlService.scala
+++ b/notifier/src/main/scala/uk/ac/wellcome/platform/archive/notifier/services/CallbackUrlService.scala
@@ -20,10 +20,11 @@ class CallbackUrlService(contextUrl: URL)(implicit actorSystem: ActorSystem,
     callbackNotification: CallbackNotification): Future[Try[HttpResponse]] = {
     for {
       jsonString <- Future.fromTry(
-        toJson(ResponseDisplayIngest(
-          ingest = callbackNotification.payload,
-          contextUrl = contextUrl
-        ))
+        toJson(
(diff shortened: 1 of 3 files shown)


notifier/src/main/scala/uk/ac/wellcome/platform/archive/notifier/services/CallbackUrlService.scala
```diff
@@ -13,26 +13,33 @@ import uk.ac.wellcome.platform.archive.display.ResponseDisplayIngest
 import scala.concurrent.{ExecutionContext, Future}
 import scala.util.{Failure, Success, Try}
 
-class CallbackUrlService(contextURL: URL)(implicit actorSystem: ActorSystem,
+class CallbackUrlService(contextUrl: URL)(implicit actorSystem: ActorSystem,
                                           ec: ExecutionContext)
     extends Logging {
   def getHttpResponse(
     callbackNotification: CallbackNotification): Future[Try[HttpResponse]] = {
```
jamesgorrie: Is the Try type still correct?
Maybe just rusty Scala of how `for`s `yield` .

alexwlchan: Yep – the `for` is yielding a `Future`, which we then wrap in `Success/Failure` as appropriate.

kenoir: Could we not convert the `Try` to a `Future` and then `flatMap` it out to remove the nested structure?

PR #178, opened 2019-04-17
Add a couple of new PRONOM mappings to the bagger

alexwlchan:
This adds a couple of new PRONOM mappings that the bagger doesn't recognise (based on the errors in the dashboard), and makes the error a bit more descriptive when it occurs.

For wellcometrust/platform#3322.

### DIFF
--- a/bagger/src/tech_md.py
+++ b/bagger/src/tech_md.py
@@ -81,8 +81,10 @@ def remodel_file_technical_metadata(root, id_map):
         )
         format_registry_name.text = "PRONOM"  # assume this is always used
         format_registry_key = make_child(format_registry, "premis", "formatRegistryKey")
-        format_registry_key.text = mappings.PRONOM[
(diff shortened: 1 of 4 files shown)


bagger/src/tech_md.py
```diff
@@ -81,8 +85,10 @@ def remodel_file_technical_metadata(root, id_map):
...
         format_registry_name.text = "PRONOM"  # assume this is always used
         format_registry_key = make_child(format_registry, "premis", "formatRegistryKey")
-        format_registry_key.text = mappings.PRONOM[
-            format_name.text
-        ]  # allow this to raise error if missing key!
+
+        try:
+            format_registry_key.text = mappings.PRONOM[format_name.text]
+        except KeyError:
+            raise UnrecognisedPRONOMFormatError(format_name.text)
```
mklander: 👍

PR #181, opened 2019-04-25
Add a bag auditor for locating a bag root

alexwlchan:


common/src/main/scala/uk/ac/wellcome/platform/archive/common/ingests/models/BetterBagRequest.scala
```diff
@@ -0,0 +1,12 @@
+package uk.ac.wellcome.platform.archive.common.ingests.models
+
+import java.util.UUID
+
+import uk.ac.wellcome.platform.archive.common.bagit.models.BagLocation
+import uk.ac.wellcome.storage.ObjectLocation
+
+case class BetterBagRequest(
```
kenoir: I think there is some overlap of `bagRoot` with the existing `BagLocation`. Maybe we could use this as an opportunity to review that class?

I don't think it blocks deployment of this though as we do need an intermediate service that "discovers" the bag route and aspirational path.

kenoir:
Nice.  Let me know when you want a final review.

PR #195, opened 2019-05-02
Add locking to the replicator

alexwlchan:
Resolves wellcometrust/platform#3599.

It works! It really works! Well, mostly. It exposes a bit of inflexibility in the Dynamo locking service that I’ll port back to the storage repo, along with the ability to record the history of locks on an in-memory LockDao (so you can find out how many locks were acquired).

But I think the rough shape of it is broadly correct. Thoughts?


bag_replicator/src/main/scala/uk/ac/wellcome/platform/archive/bagreplicator/services/BagReplicatorWorker.scala
```diff
@@ -38,26 +50,50 @@ class BagReplicatorWorker(
...
+          payload.copy(
+            bagRootLocation = replicationSummary.summary.destination
+          )
+        )
+      } yield toResult(replicationSummary)
+    }.map {
+      case Right(result) => result
+      case Left(failedLockingServiceOp) =>
+        warn(s"Unable to lock successfully: $failedLockingServiceOp")
+        NonDeterministicFailure(
```
kenoir: 👏 Was looking out for this!

alexwlchan:
Still to do:

* [ ] A crash in AlpakkaSQSWorker causes the app to hang, if it's really unrecoverable it should stop the entire application
* [ ] Add IAM permissions for changing message visibility when we return a NonDeterministicFailure
* [ ] Make the retry timeout configurable somehow – by default AlpakkaSQSWorker sets the message timeout to 0, which is precisely the wrong thing here! We want to send the message back onto the queue and leave it there for a while

PR #198, opened 2019-05-07
Add code for reading/writing a fetch file in-memory

alexwlchan:


common/src/main/scala/uk/ac/wellcome/platform/archive/common/bagit/parsers/FetchContents.scala
```diff
@@ -0,0 +1,69 @@
...
+  *     percent encoded
+  *   - `length` is the number of octets in the file, or "-" if unspecified
+  *   - `filename` is the path to the file.  Line break characters (LR, CF, LRCF)
+  *     and *only* those characters must be percent-encoded.
+  *
+  */
+object FetchContents {
+
+  val FETCH_LINE_REGEX: Regex = new Regex(
+    "(.*)[ \t]+(\\d*|-)[ \t]+(.*)", "url", "length", "filepath"
```
jamesgorrie: Worth leaving an example of what this might be?

jamesgorrie: Nevermind, you give it above.

PR #200, opened 2019-05-08
Use a URL, not a URI, for FetchEntry

alexwlchan:
This matches the spec slightly better, and lets me add tests/handling code for percent encoding of whitespace in URLs.

### DIFF
--- a/common/src/main/scala/uk/ac/wellcome/platform/archive/common/bagit/parsers/FetchContents.scala
+++ b/common/src/main/scala/uk/ac/wellcome/platform/archive/common/bagit/parsers/FetchContents.scala
@@ -47,7 +47,7 @@ object FetchContents {
         FETCH_LINE_REGEX.findFirstMatchIn(line) match {
           case Some(m) =>
             FetchEntry(
-              url = new URL(m.group("url")),
+              url = decodeURL(m.group("url")),
               length = decodeLength(m.group("length")),
               filepath = decodeFilepath(m.group("filepath"))
             )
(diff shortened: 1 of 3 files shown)


common/src/main/scala/uk/ac/wellcome/platform/archive/common/bagit/parsers/FetchContents.scala
```diff
@@ -63,10 +63,16 @@ object FetchContents {
   def write(entries: Seq[FetchEntry]): String =
     entries
       .map { e =>
-        s"${e.url} ${encodeLength(e.length)} ${encodeFilepath(e.filepath)}"
+        s"${encodeURL(e.url)} ${encodeLength(e.length)} ${encodeFilepath(e.filepath)}"
       }
       .mkString("\n")
 
+  private def encodeURL(url: URL): String =
+    url.toString.replaceAll(" ", "%20").replaceAll("\t", "%09")
```
jamesgorrie: Is there not a generic "en / decode all URL characters" - or is there a reason we don't use it?

alexwlchan: Those methods exist, but the spec is a bit unclear about whether you should encode all the characters or just some of them (although I don’t think it matters).

alexwlchan:
Closing as URI is arguably more correct; it supports `s3://` style URLs and gives us the code to decode them.

PR #203, opened 2019-05-13
Given an ingest ID, date and external ID, assign a version

alexwlchan:
There’s a Dao and all the logic for interacting it; writing the DynamoDB Dao we want to use will be a separate patch.


alexwlchan:
> Nitpick question over naming: is this the `IngestVersionManagerDao` or a `Bag IngestVersionManagerDao`?

errrrrr

PR #210, opened 2019-05-21
Use the new storage/messaging libraries

alexwlchan:
Closes https://github.com/wellcometrust/platform/issues/3629

Now we can run lots of stuff in-memory, so it's faster and easier to debug. Major changes:

* Common lib is mostly sans-IO, including the IngestUpdater and OperationPublisher
* IngestTracker has a trait underneath it, so we can use an in-memory implementation in the tests

And we run less containers in CI:

* bags_api: no S3 or DynamoDB
* bag_auditor: no SNS
* bag_register: no S3, SNS or DynamoDB
* bag_replicator: no SNS
* bag_unpacker: no SNS
* bag_verifier: no SNS
* common: no SNS
* ingests_api: no SNS or DynamoDB
* ingests_common: no SNS
* ingests: no SNS or DynamoDB
* notifier: no SNS or DynamoDB

Rough edges are being documented in https://github.com/wellcometrust/platform/issues/3630

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,7 +16,7 @@ SBT_APPS = notifier \
            bag_verifier \
            bag_unpacker \
            bag_auditor
-
+SBT_NO_DOCKER_APPS =
 
 SBT_DOCKER_LIBRARIES    = common ingests_common
 SBT_NO_DOCKER_LIBRARIES = bags_common display
--- a/api/Makefile
+++ b/api/Makefile
@@ -4,8 +4,8 @@ PROJECT_ID = storage
 
 STACK_ROOT 	= api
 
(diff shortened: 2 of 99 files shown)


alexwlchan:
\*snorts*

Apparently the replicator tests are just TOO FAST. So the locking concurrency checks can't be triggered.

One run isn’t conclusive, but the tests do look to be a few minutes faster than a comparable run elsewhere, yay.

alexwlchan:
Okay, so it’s not that they’re too fast – but refactoring `processMessage` to return a Try instead of a Future seems to have broken the concurrency.

Evening musings here: https://stackoverflow.com/q/56245298/1558022

PR #212, opened 2019-05-23
Check the length of entries referenced in fetch files

alexwlchan:
The fetch file tells us the length of the object it points to, but right now we don't check it anywhere.

This patch modifies the VerifiableLocation to carry the length specified by the fetch entry (if any), and the S3ObjectVerifier to check the length matches (if present).

You can’t get the length from a plain `S3ObjectInputStream`, so I’m wrapping it in our own class `S3ObjectStream`, which includes the Content-Length from the S3 headers.

Along the way, I discovered that those sizes can be large than Int.MaxValue, so I swapped it out for Long and added a test that we can parse fetch entries that point to files with large sizes.

Closes wellcometrust/platform#3638.

### DIFF
--- a/bag_verifier/src/test/scala/uk/ac/wellcome/platform/archive/bagverifier/services/S3ObjectVerifierTest.scala
+++ b/bag_verifier/src/test/scala/uk/ac/wellcome/platform/archive/bagverifier/services/S3ObjectVerifierTest.scala
@@ -70,7 +70,7 @@ class S3ObjectVerifierTest
     withLocalS3Bucket { bucket =>
       val contentHashingAlgorithm = MD5
       val contentString = "HelloWorld"
-      // sha256("HelloWorld")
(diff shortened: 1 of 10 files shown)


bag_verifier/src/test/scala/uk/ac/wellcome/platform/archive/bagverifier/services/S3ObjectVerifierTest.scala
```diff
@@ -72,25 +70,87 @@ class S3ObjectVerifierTest
     }
   }
 
+  it("fails if the checksum is correct but the expected length is wrong") {
+    withLocalS3Bucket { bucket =>
+      val contentHashingAlgorithm = MD5
+      val contentString = "HelloWorld"
+      // md5("HelloWorld")
```
jamesgorrie: hi

PR #216, opened 2019-05-31
Simplify interface to StorageManifestDao, hiding update

kenoir:


common/src/main/scala/uk/ac/wellcome/platform/archive/common/storage/services/StorageManifestDao.scala
```diff
@@ -2,35 +2,22 @@ package uk.ac.wellcome.platform.archive.common.storage.services
...
+  vhs: VersionedHybridStore[String, StorageManifest, EmptyMetadata]
 ) {
+  def get(id: BagId): Either[ReadError, StorageManifest] = vhs.get(s"$id")
+
+  def put(storageManifest: StorageManifest): Either[StorageError, Unit] = {
 
-  def get(id: BagId): Either[ReadError, StorageManifest] =
-    underlying.get(id = id.toString)
+    val ifNotExisting = (storageManifest, EmptyMetadata())
+    val ifExisting = (o: StorageManifest, m: EmptyMetadata) => (o, m)
```
alexwlchan: Since this will discard the update, is it better to say we should never call “put()” on an already-stored manifest? And maybe throw if we try to store something different?

common/src/test/scala/uk/ac/wellcome/platform/archive/common/fixtures/StorageManifestVHSFixture.scala
```diff
@@ -31,11 +31,7 @@ trait StorageManifestVHSFixture extends EitherValues {
   def storeSingleManifest(
     vhs: StorageManifestDao,
     storageManifest: StorageManifest): Either[StorageError, Unit] =
-    vhs.update(
-      ifNotExisting = storageManifest
-    )(
-      ifExisting = _ => throw new RuntimeException("VHS should be empty!")
-    )
+    vhs.put(storageManifest)
```
alexwlchan: Honestly now I’d just ditch this helper and in-line this everywhere, it’s shorter!

Or assert that it succeeds as part of this helper.

kenoir: Same as above - i'd like to keep these interfaces in place for now.

common/src/main/scala/uk/ac/wellcome/platform/archive/common/storage/services/StorageManifestDao.scala
kenoir: I'm not sure what the interface will be yet - i'd like to keep the `StorageManifestDao` around and functionally identical - but just simplify things to make it easier to implement multiple versions.

PR #219, opened 2019-06-04
Split out the root finding from the bag auditor, so we do it pre-verification

alexwlchan:
Closes https://github.com/wellcometrust/platform/issues/3644


alexwlchan:
I'm heading out now, but to remind myself for the morning: the problem is that the bag_root_finder payload gets passed through to the bag_auditor, and it's in the wrong format right now.

alexwlchan:
^^ this works now

common/src/test/scala/uk/ac/wellcome/platform/archive/common/generators/PayloadGenerators.scala
```diff
@@ -78,10 +78,21 @@ trait PayloadGenerators
   def createEnrichedBagInformationPayload: EnrichedBagInformationPayload =
     createEnrichedBagInformationPayload()
 
+  // TODO: Just pass the context directly.
```
alexwlchan: This is addressed in #226.

PR #222, opened 2019-06-19
Pass around Json objects for payloads

alexwlchan:
We want two things from the internal messaging:

* Type safety inside the apps
* Apps have to pass along information they don’t actually care about (e.g. the unpacker doesn't need to know the ingest date, but downstream apps do)

I did a bunch of poking, and I’m musing about this approach.

* Pass around Json objects at the edges (so extra information is handled transparently)
* Convert to a concrete case class as the first step (so we get type safety)
* Edit the Json with the data we want to add at each step

Thoughts, @kenoir?

### DIFF
--- a/bag_unpacker/src/main/scala/uk/ac/wellcome/platform/archive/bagunpacker/models/BagUnpackerPayload.scala
+++ b/bag_unpacker/src/main/scala/uk/ac/wellcome/platform/archive/bagunpacker/models/BagUnpackerPayload.scala
@@ -1,0 +1,11 @@
+package uk.ac.wellcome.platform.archive.bagunpacker.models
+
+import uk.ac.wellcome.platform.archive.common.ingests.models.IngestID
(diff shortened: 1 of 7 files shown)


alexwlchan:
Doing some more experiments with implicits on `Json`, I’m more inclined to do something like this:

```scala
implicit class PayloadJson(val underlying: Json) {
  def ingestId: Either[DecoderFailure, IngestID] = ...

  def bagRootLocation: Either[DecoderFailure, ObjectLocation] = ...
  def withBagRootLocation: Json = ...
}
```

and then in the apps, you’d have something like

```
def processMessage(json: Json) =
  for {
    ingestId <- json.ingestId
    bagRootLocation <- json.bagRootLocation

    ... // do stuff //

    outgoingJson <- json.withBagRootLocation(root).withExternalIdentifier(externalId)
    _ <- outgoingPublisher.send(outgoingJson)
  }
```

PR #227, opened 2019-06-21
Bag auditor checks ingest type before assigning a version

alexwlchan:
Closes wellcometrust/platform#3657, and possibly wellcometrust/platform#3499?

It includes some new user-facing messages in the ingests response:

> Unable to find an external identifier
>
> This bag has never been ingested before, but was sent with ingestType update
>
> This bag has already been ingested, but was sent with ingestType create
>
> Another version of this bag was ingested at ${e.stored}, which is newer than the current ingest ${e.request}

I discovered wellcometrust/platform#3693 while working on this.

### DIFF
--- a/bag_auditor/src/main/scala/uk/ac/wellcome/platform/storage/bagauditor/versioning/VersionPicker.scala
+++ b/bag_auditor/src/main/scala/uk/ac/wellcome/platform/storage/bagauditor/versioning/VersionPicker.scala
@@ -5,19 +5,12 @@ import java.util.UUID
 
 import cats.implicits._
 import uk.ac.wellcome.platform.archive.common.bagit.models.ExternalIdentifier
-import uk.ac.wellcome.platform.archive.common.ingests.models.{
-  CreateIngestType,
-  IngestID,
-  IngestType,
-  UpdateIngestType
-}
+import uk.ac.wellcome.platform.archive.common.ingests.models.{CreateIngestType, IngestID, IngestType, UpdateIngestType}
 import uk.ac.wellcome.platform.archive.common.versioning.IngestVersionManager
+import uk.ac.wellcome.platform.storage.bagauditor.models.{IngestTypeCreateForExistingBag, IngestTypeUpdateForNewBag, InternalVersionPickerError, VersionPickerError}
 import uk.ac.wellcome.storage.{FailedProcess, LockDao, LockingService}
 
-import scala.util.Try
-
-class IllegalVersionAssignment(message: String)
-    extends RuntimeException(message)
+import scala.util.{Failure, Success, Try}
 
 class VersionPicker(
   lockingService: LockingService[Int, Try, LockDao[String, UUID]],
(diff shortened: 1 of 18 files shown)


jtweed:
What about something like this for the messages:

- Unable to ingest bag. An external identifier was not found in the bag info.
- Unable to ingest bag. Another version of this bag was ingested at $PREVIOUS_INGEST_DATE, which is newer than the current ingest $CURRENT_INGEST_DATE.
- Unable to ingest new bag. A bag with the supplied external identifier already exists in this space.
- Unable to update existing bag. A bag with the supplied external identifier does not exist in this space.

alexwlchan:
@jtweed I’m picking nits, but “external identifier was not found in the bag info” implies that were able to find the bag info. This step could also fail if the bag info is just missing. Do we care about that distinction?

jtweed:
If the bag info itself is missing, that would ideally be a distinct error. If we can’t tell the difference, then I think this message is still correct, as we couldn’t find an id in the bag info, as there was no bag info.

alexwlchan:
> If the bag info itself is missing, that would ideally be a distinct error. If we can’t tell the difference, then I think this message is still correct, as we couldn’t find an id in the bag info, as there was no bag info.

Not currently; I'll change the error messages here and open a new ticket to make it a distinct error, because I expect it'll need a moderate refactor.

bag_auditor/src/main/scala/uk/ac/wellcome/platform/storage/bagauditor/versioning/VersionPicker.scala
```diff
@@ -3,57 +3,87 @@ package uk.ac.wellcome.platform.storage.bagauditor.versioning
...
       }
-      .map {
-        case Right(version) =>
-          checkVersionIsAllowed(ingestType, assignedVersion = version)
-          version
 
-        case Left(FailedProcess(_, err)) => throw err
-        case Left(err)                   => throw new RuntimeException(s"Locking error: $err")
-      }
+    // This is a double Either: the outer Either defines the result of the locking
```
kenoir: https://typelevel.org/cats/datatypes/nested.html

Might come in handy next time!

bag_auditor/src/main/scala/uk/ac/wellcome/platform/storage/bagauditor/versioning/VersionPicker.scala
```diff
@@ -3,57 +3,87 @@ package uk.ac.wellcome.platform.storage.bagauditor.versioning
...
+      Left(IngestTypeCreateForExistingBag())
     } else if (ingestType == UpdateIngestType && assignedVersion == 1) {
-      throw new IllegalVersionAssignment(
-        "Ingest type 'update' is not allowed unless a bag already exists")
+      Left(IngestTypeUpdateForNewBag())
+    } else {
+      Right(assignedVersion)
+    }
+
+  // Annoyingly, cats doesn't seem to provide an Implicit for the Id monad, so we have
```
kenoir: I think it does? Does `import cats._` not work?

bag_auditor/src/main/scala/uk/ac/wellcome/platform/storage/bagauditor/versioning/VersionPicker.scala
alexwlchan: Next time!

PR #246, opened 2019-07-05
First pass at including poster images in bags and METS

tomcrane:
Addresses https://github.com/wellcometrust/platform/issues/3718

### DIFF
--- a/bagger/src/bag_assembly.py
+++ b/bagger/src/bag_assembly.py
@@ -23,12 +23,11 @@ def cleanup(bag_details):
     if os.path.isdir(directory):
         shutil.rmtree(directory)
 
-    zipfile = bag_details["zip_file_path"]
-
-    logging.debug("deleting %s", zipfile)
-
-    if os.path.isfile(zipfile):
-        os.remove(zipfile)
+    zipfile = bag_details.get("zip_file_path", None)
+    if zipfile is not None:
+        logging.debug("deleting %s", zipfile)
+        if os.path.isfile(zipfile):
+            os.remove(zipfile)
 
 
 def prepare_bag_dir(b_number):
--- a/bagger/src/files.py
+++ b/bagger/src/files.py
@@ -10,6 +10,7 @@ import settings
 import bag_assembly
 import dlcs
 import storage
+import tech_md
 from xml_help import expand, namespaces
 
 import boto3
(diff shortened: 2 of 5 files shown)


tomcrane:
btw I didn't want to add any image library dependencies just to find the width and height of the poster image, hence https://github.com/wellcometrust/storage-service/pull/246/files#diff-6e9a183acdd34cfd279c0485e722c9cdR5 

But that can go if not wanted.

bagger/src/tech_md.py
```diff
@@ -92,3 +98,39 @@ def remodel_file_technical_metadata(root, id_map):
...
+    # structMaps (yet), as that would be just inventing something that needs to
+    # go in Archivematica
+
+    # data we are interested in
+    head, tail = os.path.split(destination)
+    file_name = tail
+    file_size = os.path.getsize(destination)
+    width, height = image_size.get_image_size(destination)
+
+    amds = root.find("./mets:amdSec[@ID='AMD']", namespaces)
```
kenoir: Breaking things up a bit here would make this easier to read.

bagger/src/bag_assembly.py
```diff
@@ -23,12 +23,11 @@ def cleanup(bag_details):
...
-
-    logging.debug("deleting %s", zipfile)
-
-    if os.path.isfile(zipfile):
-        os.remove(zipfile)
+    zipfile = bag_details.get("zip_file_path", None)
+    if zipfile is not None:
+        logging.debug("deleting %s", zipfile)
+        if os.path.isfile(zipfile):
+            os.remove(zipfile)
```
alexwlchan: Python convention is to try the thing and catch the exception, rather than all these if checks.

i.e.,

```python
try:
    zipfile = bag_details["zip_file_path"]
    logging.debug("deleting %s", zipfile)
    try:
        os.remove(zipfile)
    except FileDoesNotExistError:
        pass
except KeyError:
    pass
```

But I don't think it makes a massive difference here.

bagger/src/files.py
```diff
@@ -20,6 +21,10 @@
 ALTO_KEYS = set()
 OBJECT_KEYS = set()
 
+# For bagging posters when running on AWS
+POSTER_CANDIDATES = [".mpg", ".mpeg", ".mp3", ".wav", ".mp4"]
```
alexwlchan: tiny optimisation: make this a set so lookups are O(1) rather than linear.
```suggestion
POSTER_CANDIDATES = {".mpg", ".mpeg", ".mp3", ".wav", ".mp4"}
```

bagger/src/tech_md.py
alexwlchan: Seconded. Plus a brief comment about what it’s doing to the METS structure – what XML are you trying to create – would be helpful.

PR #239, opened 2019-07-02
Add DynamoDB for bag_id_lookup table

kenoir:

PR #249, opened 2019-07-09
Get the pipeline back into a working state

alexwlchan:


alexwlchan:
This is mostly stuff Robert and I worked on together (the salvaged remains of #239) plus enough to get the tests passing again, so merging by fiat.

PR #251, opened 2019-07-10
Create RFC-compliant storage manifests

alexwlchan:
So if a bag has a file in a different version, we'll resolve it to create an RFC-compliant path.

Still to do:

- [x] Actually use this in the bag register
- [x] Modify the display models to match


alexwlchan:
This is deployed and working!

alexwlchan:
Hmm. I want to think about the consistency of version information in the API.

e.g. we use "v2" in path but you request `?version=2`. Should we use "v2" everywhere externally?

bag_register/src/test/scala/uk/ac/wellcome/platform/archive/bag_register/fixtures/BagRegisterFixtures.scala
```diff
@@ -101,4 +102,21 @@ trait BagRegisterFixtures
...
     }
+
+  // The bag register inspects the paths to a bag's entries to
+  // check they are in the correct format post-replicator,
+  // hence the version directory.
+  def withBag[R](bucket: Bucket,
+                 bagInfo: BagInfo,
+                 externalIdentifier: ExternalIdentifier,
+                 space: StorageSpace,
+                 version: Int)(testWith: TestWith[ObjectLocation, R]): R =
```
kenoir: 👍

common/src/test/scala/uk/ac/wellcome/platform/archive/common/generators/StorageManifestGenerators.scala
```diff
@@ -38,21 +37,21 @@ trait StorageManifestGenerators
...
       ),
-      tagManifest = BagManifest(
+      tagManifest = FileManifest(
         checksumAlgorithm,
         files = Seq(
           createBagFile,
           createBagFile,
           createBagFile
-        )
+        ).map { StorageManifestFile.apply }
```
kenoir: As a style thing I would not make this in-line as I reckon it's easier to read.

alexwlchan: oh hold on, that shouldn't be there

alexwlchan: This was temporary code to keep that generator compiling while I was fiddling elsewhere, but I meant to take it out

PR #257, opened 2019-07-12
Improve the user-facing error messages from the unpacker

alexwlchan:
This adds two new error messages to the unpacker:

* If we go to unpack a bag from S3 and discover it's in a bucket we can't read, we send the message

    > Unpacking failed - Access denied while trying to read s3://bucket/archive.tar.gz

* If we go to unpack a bag and discover it's not in tar.gz format, we send the message

    > Unpacking failed - Error trying to unpack the archive at bucket/archive.tar.gz - is it the correct format?

This is one of the things @jtweed flagged in testing. We need to be a bit more consistent about when we use S3 URIs in user-facing messages, but that's a separate patch.

### DIFF
--- a/bag_unpacker/src/main/scala/uk/ac/wellcome/platform/archive/bagunpacker/services/Unpacker.scala
+++ b/bag_unpacker/src/main/scala/uk/ac/wellcome/platform/archive/bagunpacker/services/Unpacker.scala
@@ -66,8 +66,8 @@ trait Unpacker {
     }
   }
 
-  private def buildMessageFor(srcLocation: ObjectLocation,
-                              error: UnpackerError): Option[String] =
+  protected def buildMessageFor(srcLocation: ObjectLocation,
+                                error: UnpackerError): Option[String] =
     error match {
       case UnpackerStorageError(_: DoesNotExistError) =>
         Some(s"There is no archive at $srcLocation")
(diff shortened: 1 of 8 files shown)


bag_unpacker/src/main/scala/uk/ac/wellcome/platform/archive/bagunpacker/services/Unpacker.scala
```diff
@@ -66,12 +66,16 @@ trait Unpacker {
...
-                              error: UnpackerError): Option[String] =
+  protected def buildMessageFor(srcLocation: ObjectLocation,
+                                error: UnpackerError): Option[String] =
     error match {
       case UnpackerStorageError(_: DoesNotExistError) =>
         Some(s"There is no archive at $srcLocation")
 
+      case UnpackerUnarchiverError(_) =>
+        Some(
+          s"Error trying to unpack the archive at $srcLocation - is it the correct format?")
```
jtweed: `- not a valid .tar.gz`? or can we not necessarily tell and it only could be that? If so, then your suggestion is 👍

alexwlchan: We can’t really tell – in theory the unpacker supports any archive + compression type format.

PR #262, opened 2019-07-15
Improve the unpacker user-facing messages some more

alexwlchan:
New messages:

* Not all strings are valid S3 bucket names (e.g. bucket names must be lowercase), so if the SDK complains we pass that up to the user:

    > Unpacking failed - BUKKIT is not a valid S3 bucket name

* If you tell us to look in a bucket that flat out doesn't exist, previously we'd send a slightly generic "Could not read archive", now we send

    > Unpacking failed - There is no S3 bucket mygreatbukkit

* And now if the bucket exists, but the key doesn't, we send:

    > Unpacking failed - There is no archive at s3://mygreatbukkit/archive.tar.gz

  Previously we wouldn't add the `s3://` prefix; now we're consistent.

Closes https://github.com/wellcometrust/platform/issues/3735.

We can give a better error if the source bucket is in a different region (https://github.com/wellcometrust/platform/issues/3736), but I haven't found a way to test it reliably.

### DIFF
--- a/bag_unpacker/src/main/scala/uk/ac/wellcome/platform/archive/bagunpacker/services/s3/S3Unpacker.scala
+++ b/bag_unpacker/src/main/scala/uk/ac/wellcome/platform/archive/bagunpacker/services/s3/S3Unpacker.scala
@@ -14,7 +14,12 @@ import uk.ac.wellcome.storage.streaming.{
   InputStreamWithLength,
   InputStreamWithLengthAndMetadata
 }
-import uk.ac.wellcome.storage.{DoesNotExistError, ObjectLocation, StorageError, StoreReadError}
+import uk.ac.wellcome.storage.{
+  DoesNotExistError,
+  ObjectLocation,
+  StorageError,
+  StoreReadError
+}
 
 class S3Unpacker()(implicit s3Client: AmazonS3) extends Unpacker {
   private val s3StreamStore = new S3StreamStore()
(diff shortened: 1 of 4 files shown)


bag_unpacker/src/test/scala/uk/ac/wellcome/platform/archive/bagunpacker/UnpackerFeatureTest.scala
```diff
@@ -101,7 +103,7 @@ class UnpackerFeatureTest
                 ingestUpdates.tail.head.asInstanceOf[IngestStatusUpdate]
               ingestFailed.status shouldBe Ingest.Failed
               ingestFailed.events.head.description shouldBe
-                s"Unpacker failed - There is no archive at ${payload.sourceLocation}"
+                s"Unpacker failed - There is no S3 bucket ${sourceLocation.namespace}"
```
kenoir: I wonder whether we should be non cloud provider specific at this level.

bag_unpacker/src/main/scala/uk/ac/wellcome/platform/archive/bagunpacker/services/s3/S3Unpacker.scala
```diff
@@ -40,6 +45,17 @@ class S3Unpacker()(implicit s3Client: AmazonS3) extends Unpacker {
           if exc.getMessage.startsWith("Access Denied") =>
         Some(s"Access denied while trying to read s3://$srcLocation")
 
+      case UnpackerStorageError(StoreReadError(exc: AmazonS3Exception))
```
kenoir: As below - we should think about how to do this in a way that is not specific to cloud provider.

alexwlchan: Let’s talk about it when I’m in this afternoon, because I was hoping it was already that way!

In line with the way we've done the scala-storage classes, there's a base `Unpacker` trait, and then implementations of that with different backends (memory, S3). I think it's okay for implementations to include cloud provider-specific messages.

If the unpacker supports multiple backends, there'll need to be a way for a user to say “unpack this bag from S3” or “unpack this bag from Azure” – so knowing the unpacker looked in a different provider might be useful!

PR #263, opened 2019-07-16
Tidy up the implementation/interface of StorageManifestDao

alexwlchan:
Refactoring for https://github.com/wellcometrust/platform/issues/3731

This replaces StorageManifestDao with a trait of the same name, and in-memory/DynamoDB implementations. Next we'll have to add a `listVersionsOf(bagId: BagId)` method to supply the `/versions` endpoint.

This also:

* Adds a bit of testing around the Dynamo implementation, which was previously mostly untested
* We no longer have to write a map `alex -> test` into DynamoDB to satisfy the DynamoFormat requirements


common/src/main/scala/uk/ac/wellcome/platform/archive/common/storage/services/memory/MemoryStorageManifestDao.scala
```diff
@@ -0,0 +1,16 @@
+package uk.ac.wellcome.platform.archive.common.storage.services.memory
```
kenoir: Do you think these memory versions should be in the test namespace? I think it makes sense for them to **not be** in the library, but here it feels like they should never be used except in tests.

alexwlchan: I have no strong feelings, but it feels a bit odd to be inconsistent across the two libraries. Either memory implementations are for testing or they’re not.

PR #265, opened 2019-07-16
Add a minimal /versions endpoint to the bags API

alexwlchan:
Following the model defined in https://github.com/wellcometrust/platform/pull/3733, for https://github.com/wellcometrust/platform/issues/3731

This is missing a couple of pieces:

* The manifests don't have URLs on them; you have to know what to do with the version
* There's no `"latest": true` flag yet

Both are small frills that can be tackled in a separate PR (or dropped).

Plus better test coverage of the bags API. (Tests for internal server errors are back, and they're logging better info to boot.)

### DIFF
--- a/api/bags_api/src/main/scala/uk/ac/wellcome/platform/storage/bags/api/Router.scala
+++ b/api/bags_api/src/main/scala/uk/ac/wellcome/platform/storage/bags/api/Router.scala
@@ -93,6 +93,21 @@ class Router(storageManifestDao: StorageManifestDao, contextURL: URL)(
             }
           }
         }
+      } ~ path(Segment / Segment / "versions") { (space, externalIdentifier) =>
+        val bagId = BagId(
+          space = StorageSpace(space),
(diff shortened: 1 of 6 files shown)


api/bags_api/src/main/scala/uk/ac/wellcome/platform/storage/bags/api/Router.scala
```diff
@@ -48,18 +51,16 @@ class Router(storageManifestDao: StorageManifestDao, contextURL: URL)(
...
-                  versionRegex.findFirstMatchIn(versionString) match {
-                    case Some(regexMatch) =>
-                      storageManifestDao.get(
-                        bagId,
-                        version = regexMatch.group("version").toInt)
-                    case None => Left(NoVersionExistsError())
-                  }
-                case None => storageManifestDao.getLatest(bagId)
-              }
+            val result = parseVersion(maybeVersion) match {
```
kenoir: 👍

PR #266, opened 2019-07-17
Tidying in the internal models

alexwlchan:
We're passing around parameters we don't really need, so ditch them.


bag_auditor/src/main/scala/uk/ac/wellcome/platform/storage/bagauditor/models/AuditSummary.scala
```diff
@@ -3,26 +3,19 @@ package uk.ac.wellcome.platform.storage.bagauditor.models
...
 import uk.ac.wellcome.platform.archive.common.operation.models.Summary
-import uk.ac.wellcome.platform.archive.common.storage.models.StorageSpace
-import uk.ac.wellcome.storage.ObjectLocation
 
 sealed trait AuditSummary extends Summary {
-  val root: ObjectLocation
-  val space: StorageSpace
-  val startTime: Instant
+  val endTime: Instant
+  override val maybeEndTime: Option[Instant] = Some(endTime)
```
kenoir: Why both these fields?

PR #269, opened 2019-07-18
Add support for getting bag by version

helenst:
#3726

### DIFF
--- a/python_client/src/wellcome_storage_service/__init__.py
+++ b/python_client/src/wellcome_storage_service/__init__.py
@@ -131,16 +131,20 @@ class StorageServiceClient:
 
     @needs_token
     @check_api_resp
-    def get_bag(self, space_id, source_id):
+    def get_bag(self, space_id, source_id, version=None):
         """
         Get an individual bag.
         """
-        bags_api_url = self.api_url + "/bags/%s/%s" % (space_id, source_id)
+        bags_url_suffix = '%s/%s' % (space_id, source_id)
+        if version:
+            bags_url_suffix += '?version=%s' % version
+
+        bags_api_url = self.api_url + "/bags/%s" % bags_url_suffix
         resp = self.sess.get(bags_api_url)
 
         if resp.status_code == 404:
             raise BagNotFound(
-                "Bags API returned 404 for bag %s/%s" % (space_id, source_id)
+                "Bags API returned 404 for bag %s" % bags_url_suffix
             )
         else:
             return resp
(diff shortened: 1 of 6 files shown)


python_client/tests/test_bags.py
```diff
@@ -12,6 +12,13 @@ def test_can_get_bag(client):
     assert len(resp["manifest"]["files"]) == 3
 
 
+def test_can_get_bag_version(client):
+    resp = client.get_bag(space_id="alex-testing", source_id="b12345", version="v1")
+    assert resp["space"]["id"] == "alex-testing"
+    assert resp["id"] == "alex-testing/b12345"
+    assert len(resp["manifest"]["files"]) == 3
```
alexwlchan: What about a test for `v2` of this bag, to check that if you request a version that doesn't exist, it returns correctly?

And maybe a test for a `v3` that does exist, and asserting it returns the correct version?

helenst: I've added a test for missing bag (since it does trigger a different code path here). I'm not sure what a test for v3 would add given that the client has no logic for returning the correct version and relies on the server to do so, but it's possible I'm missing something here :)

alexwlchan: If the storage service only knows about `v1`, you'll get an identical response for `/bag` and `/bag?version=v1`. The idea of a test for v3 is to make sure the client is actually adding that `?version` parameter.

helenst: That makes sense :-) Ready for re-review!

python_client/src/wellcome_storage_service/__init__.py
```diff
@@ -131,16 +131,22 @@ def create_s3_ingest(
...
+        resp = self.sess.get(bags_api_url, params=params)
 
         if resp.status_code == 404:
-            raise BagNotFound(
-                "Bags API returned 404 for bag %s/%s" % (space_id, source_id)
-            )
+            err = "Bags API returned 404 for bag %s" % bags_url_suffix
+            if version:
+                err += " with version %s" % version
+            raise BagNotFound(err)
```
alexwlchan: 👌

python_client/tests/test_bags.py
```diff
@@ -12,7 +12,28 @@ def test_can_get_bag(client):
...
+    resp = client.get_bag(space_id="alex-testing", source_id="b12345", version="v1")
+    assert resp["space"]["id"] == "alex-testing"
+    assert resp["id"] == "alex-testing/b12345"
+    assert resp["version"] == "v1"
+    assert len(resp["manifest"]["files"]) == 3
+
+
+def test_gets_correct_bag_version(client):
+    resp = client.get_bag(space_id="alex-testing", source_id="b12345", version="v3")
+    assert resp["version"] == "v3"
```
alexwlchan: Nice!

PR #270, opened 2019-07-22
Adds changes to use the new AzureSSO roles

kenoir:


build.sbt
```diff
@@ -44,10 +47,23 @@ def setupProject(
     .settings(libraryDependencies ++= externalDependencies)
 }
 
+// In order to access our libraries in S3 we need to set the following:
+
+s3CredentialsProvider := { _ =>
+  val builder = new STSAssumeRoleSessionCredentialsProvider.Builder(
+    roleArn = "arn:aws:iam::760097843905:role/platform-read_only",
```
kenoir: This allows you to provide only the default credentials which allow you to switch into the read_only role.

PR #280, opened 2019-07-26
Send the version in the notifier callback

alexwlchan:
* When the auditor assigns a version, it sends an IngestVersionUpdate to the ingests monitor, which attaches it to the ingest. This version gets shown in the ingests API and the notifier. Fixes https://github.com/wellcometrust/platform/issues/3770

* The IngestStatesTest has been split into one test for each type of update, with the event-updating test cases a single shared set.

* The notifier tests now assert on explicit JSON output, which fixes a couple of cases where we were sending null in the output

* Don't make `lastModifiedDate` an explicit field on an `Ingest`; derive it from the list of IngestEvents when we need to display it. We were never setting it previously, now it’s marginally more useful. This fell out of refactoring to catch nulls. Closes https://github.com/wellcometrust/platform/issues/3773


alexwlchan:
Even if this passes, tests are flaky because I mixed in some extra randomness for lastModifiedDate (which previously we _never_ set).

Diverting to fix #3773 first.

alexwlchan:
This is getting too big, I’m going to break it up into smaller PRs.

alexwlchan:
Closing in favour of #286.

PR #282, opened 2019-07-29
Set the last modified date on an ingest

alexwlchan:
Closes https://github.com/wellcometrust/platform/issues/3773. Spun out as a separate piece from #280.

Rather than setting the last modified date explicitly, we compute it from the latest IngestEvent date.

### DIFF
--- a/api/ingests_api/src/test/scala/uk/ac/wellcome/platform/storage/ingests/api/IngestsApiFeatureTest.scala
+++ b/api/ingests_api/src/test/scala/uk/ac/wellcome/platform/storage/ingests/api/IngestsApiFeatureTest.scala
@@ -36,7 +36,9 @@ class IngestsApiFeatureTest
     it("returns a ingest tracker when available") {
       val ingest = createIngestWith(
         createdDate = Instant.now(),
-        events = Seq(createIngestEvent, createIngestEvent).sortBy { _.createdDate }
+        events = Seq(createIngestEvent, createIngestEvent).sortBy {
+          _.createdDate
+        }
       )
 
       withConfiguredApp(initialIngests = Seq(ingest)) {
--- a/api/ingests_api/src/test/scala/uk/ac/wellcome/platform/storage/ingests/api/IngestsApiFeatureTest.scala
+++ b/api/ingests_api/src/test/scala/uk/ac/wellcome/platform/storage/ingests/api/IngestsApiFeatureTest.scala
@@ -36,7 +36,7 @@ class IngestsApiFeatureTest
     it("returns a ingest tracker when available") {
       val ingest = createIngestWith(
         createdDate = Instant.now(),
-        events = Seq(createIngestEvent, createIngestEvent)
+        events = Seq(createIngestEvent, createIngestEvent).sortBy { _.createdDate }
       )
 
       withConfiguredApp(initialIngests = Seq(ingest)) {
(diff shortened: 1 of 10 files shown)


api/ingests_api/src/test/scala/uk/ac/wellcome/platform/storage/ingests/api/IngestsApiFeatureTest.scala
```diff
@@ -90,6 +92,7 @@ class IngestsApiFeatureTest
                      |    }
                      |  },
                      |  "createdDate": "${ingest.createdDate}",
+                     |  "lastModifiedDate": "${ingest.lastModifiedDate.get}",
```
jamesgorrie: We're 100% sure there's something in the `Option`?

alexwlchan: Yep – because `lastModifiedDate` is populated if and only if the `events` field is non-empty, which it is.

Also, this is test code – if this doesn’t contain a `lastModifiedDate`, that’s something interesting we’d like to know about!

common/src/test/scala/uk/ac/wellcome/platform/archive/common/generators/IngestGenerators.scala
```diff
@@ -45,7 +45,17 @@ trait IngestGenerators extends BagIdGenerators {
...
+
+  def createIngestEventWith(
+    description: String = randomAlphanumeric,
+    createdDate: Instant =
+      Instant.now().plusSeconds(randomInt(from = 0, to = 30))
+  ): IngestEvent =
+    IngestEvent(
+      description = description,
+      createdDate = createdDate
+    )
```
warrd: Unless I'm mistaken, won't these default parameters just get initialised once, rather than randomly at each call?

Assuming that is not what you want, maybe would make sense for  the function to take `Option[_]` instead, and default to `None`, e.g.

```scala
def createIngestEventWith(
  description: Option[String] = None,
  createdDate: Option[Instant] = None
    
): IngestEvent =
  IngestEvent(
    description = description.getOrElse(randomAlphanumeric),
    createdDate = createdDate.getOrElse(
      Instant.now.plusSeconds(randomInt(from = 0, to = 30))
    )
  )

```

alexwlchan: Perhaps surprisingly, no! (I had to double-check this – and thanks for making me check. I’ve seen what looks like the correct behaviour in the past, but never checked explicitly.)

Here's a little repro (from master, so it's just got the `randomAlphanumeric` bit):

```scala
import uk.ac.wellcome.platform.archive.common.generators.IngestGenerators
import uk.ac.wellcome.platform.archive.common.ingests.models.IngestEvent

object Experiments extends IngestGenerators {
  def x(): Seq[IngestEvent] = {
    Seq(createIngestEvent, createIngestEvent, createIngestEvent)
  }
}

println(Experiments.x())

// List(
//   IngestEvent(ImogKzTd,2019-07-29T11:09:22.099Z),
//   IngestEvent(NK1WKolY,2019-07-29T11:09:22.099Z),
//   IngestEvent(TvfYL6CJ,2019-07-29T11:09:22.099Z)
// )
```

According to the [Scala docs](https://docs.scala-lang.org/sips/completed/named-and-default-arguments.html) about default arguments:

> A method parameter with a default argument has the form `"p: T = expr"`, where `expr` is evaluated every time a method application uses the default argument

warrd: :+1: ah that is good to know that it evaluates the expression each time for default args, no what I was expecting

PR #288, opened 2019-07-29
Better error message if a bag points to something in the fetch file *and* includes a concrete file

alexwlchan:
Suppose you supply a bag with `myfile.jpg`. The file appears in two places:

* A remote location in the `fetch.txt`
* A concrete file in the bag

Previously, the verifier would reject this bag with the slightly puzzling message:

> Bag contains a file which is not referenced in the manifest: myfile.jpg

Now it explains it a bit better:

> Files referred to in the fetch.txt also appear in the bag: myfile.jpg

A concrete file or a remote location – pick one!

Plus a bunch of fixes that came from *actually* turning on fetch files in tests – the bags created by `createBag()` will now put some files in a fetch.txt, which should make the tests more robust.

Closes https://github.com/wellcometrust/platform/issues/3776


alexwlchan:
Closing in favour of #290.

PR #293, opened 2019-08-01
Collect the sizes in StorageManifestService

alexwlchan:
Closes https://github.com/wellcometrust/platform/issues/3790

The bag register will now read all the sizes of objects in a bag (either from the fetch.txt or from S3), and then include those in the StorageManifest it saves to S3.

### DIFF
--- a/common/src/main/scala/uk/ac/wellcome/platform/archive/common/storage/services/StorageManifestService.scala
+++ b/common/src/main/scala/uk/ac/wellcome/platform/archive/common/storage/services/StorageManifestService.scala
@@ -31,6 +31,7 @@ object StorageManifestService extends Logging {
   def createManifest(
     bag: Bag,
     replicaRoot: ObjectLocation,
+    sizes: Map[ObjectLocation, Long],
     space: StorageSpace,
     version: BagVersion
   ): Try[StorageManifest] = {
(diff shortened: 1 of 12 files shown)


common/src/main/scala/uk/ac/wellcome/platform/archive/common/storage/services/StorageManifestService.scala
```diff
@@ -105,65 +110,77 @@ object StorageManifestService extends Logging {
...
     *     versioned directory under the same bag root
     *
-    * This function gets a map (bag name) -> (path), relative to the bag root.
+    * The map contains data
+    *
+    *   (bag path) -> (location, size if known)
     *
     */
-  private def createNamePathMap(
+  private def createPathLocationMap(
```
warrd: I'm missing a fair bit of context, but I'm finding this method pretty nested and unwieldy (I know was already pretty big to start with!). Maybe it would be easier to read if some of the inner functionality was broken into a few smaller methods?

alexwlchan: Yep, fair comment, let me break it up a bit.

alexwlchan:
Something weird is happening with the bagger-publish task. 🙁

alexwlchan:
And the bag_register is flaky. 😭 

Merging anyway, so I can get this deployed for Intranda.

PR #307, opened 2019-08-08
Update SBT to 1.2.8

kenoir:
Use faster/more recent version of SBT - remove deprecated usage of util.Json.

This should also fix the current issue with the Travis build.


bag_unpacker/src/main/resources/logback.xml
```diff
@@ -1,7 +1,7 @@
 <configuration>
   <appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
     <encoder>
-      <pattern>%d{HH:mm:ss.SSS} [%thread] %-5level %logger{36} - %msg%n</pattern>
+      <pattern>%d{HH:mm:ss.SSS} %X{ingestId} %X{externalIdentifier} [%thread] %-5level %logger{36} - %msg%n</pattern>
```
jamesgorrie: Where are you getting there from ?

kenoir: leftover code

project/build.properties
```diff
@@ -1 +1 @@
-sbt.version=0.13.17
+sbt.version=1.2.8
```
alexwlchan: How did I not notice that 0.13.17 is older than 1.2.8?

🤦‍♀

kenoir: I think the 0.13 branch is forked from the 1.x branch and still maintained, so it's not clear if it is actually older. The `0.13` fork is the "old" one though.

alexwlchan:
Close enough.

PR #311, opened 2019-08-12
Check the tag manifests match in the bag replicator

alexwlchan:
Closes https://github.com/wellcometrust/platform/issues/3823


bag_replicator/src/test/scala/uk/ac/wellcome/platform/archive/bagreplicator/services/BagReplicatorWorkerTest.scala
```diff
@@ -327,4 +338,114 @@ class BagReplicatorWorkerTest
       }
     }
   }
+
+  describe("checks the tagmanifest-sha256.txt is correct") {
```
jamesgorrie: Is this description actually what we're doing?
Are we checking that the validation of the `tagmanifest` functions correctly?

alexwlchan: Implicitly – we've already verified that the original bag has a valid `tagmanifest-sha256.txt` by this point, and here we're verifying that the replicated bag has an identical one.

I've tweaked the description – better?

PR #315, opened 2019-08-12
Add a passthrough replication aggregator

kenoir:
In order to ensure that our archival bags is stored securely we need to replicate those bags to multiple locations.

This PR adds a "replication aggregator" service whose responsibility is to count successful replicas and pass along notifications for those that fulfil the required number & type of replicas. 

**TODO:**
- [ ] Add infrastructure for service
- [ ] Wire up service with SQS/SNS
- [ ] Complete passthrough test


replica_aggregator/src/main/resources/application.conf
```diff
@@ -0,0 +1,9 @@
+aws.sqs.queue.url=${?queue_url}
+aws.ingest.sns.topic.arn=${?ingest_topic_arn}
+aws.outgoing.sns.topic.arn=${?outgoing_topic_arn}
+aws.metrics.namespace=${?metrics_namespace}
+operation.name=${?operation_name}
+aws.locking.dynamo.tableName=${?locking_table_name}
+aws.locking.dynamo.tableIndex=${?locking_table_index}
+aws.versions.dynamo.tableName=${?versions_table_name}
+aws.versions.dynamo.tableIndex=${?versions_table_index}
```
alexwlchan: Very minor, but since this is new: could we alphabetise this list? Just makes it a tad easier to read and see if a particular flag is defined.

replica_aggregator/src/main/scala/uk/ac/wellcome/platform/storage/replica_aggregator/services/ReplicaAggregator.scala
```diff
@@ -0,0 +1,57 @@
...
+      case Left(failed) =>
+        Success(
+          IngestFailed(
+            ReplicationAggregationFailed(
+              replicationSet = ReplicationSet(replicaIdentifier, Set.empty),
+              startTime = startTime,
+              endTime = Instant.now
+            ),
+            failed.e,
+            Some("Failed to aggregate replicas!")
```
alexwlchan: We shouldn’t be exposing this as a user-facing message – this is an internal thing, isn't it?

alexwlchan:
Literally all this does right now is send a message to the ingests monitor and pop it onto the next queue.

PR #316, opened 2019-08-13
Bump the scala-storage version to 7.19.0

alexwlchan:
This should get us parallelism in the bag replications, which make them a bit faster and more efficient.

Closes https://github.com/wellcometrust/platform/issues/3727


alexwlchan:
(Failing Travis is because I accidentally pushed to master first, which triggered a build that I’ve cancelled.)

PR #319, opened 2019-08-13
Actually store results in the replicator aggregator

alexwlchan:


alexwlchan:
This is a standalone patch that we can merge – the next patch will be actually calling this code as we go throw, and setting up all the bits we need in DynamoDB and IAM and so on.

replica_aggregator/src/main/scala/uk/ac/wellcome/platform/storage/replica_aggregator/services/ReplicaAggregator.scala
```diff
@@ -0,0 +1,67 @@
...
+      //
+      //    * at least one primary replica
+      //    * at least N secondary replicas, where N is configurable
+      //
+      // For the initial, simpler case, we just handle primary replicas, so
+      // error out if passed a secondary replica.
+      //
+      result.storageLocation match {
+        case _: SecondaryStorageLocation =>
+          throw new Throwable(
```
kenoir: Though I think this is interpreted as a deterministic error - it would be nicer to return a result type here indicative of that failure mode to be interpreted elsewhere. The intention is to replace this code path soon so no need to fix now.

alexwlchan: Yes, it's deliberately rubbish!

I wanted to put *something* here so if the pipeline goes wrong we see it, but it's coming out the next time we touch this code.

replica_aggregator/src/main/scala/uk/ac/wellcome/platform/storage/replica_aggregator/models/ReplicationAggregationSummary.scala
```diff
@@ -5,7 +5,7 @@ import java.time.Instant
 import uk.ac.wellcome.platform.archive.common.operation.models.Summary
 
 sealed trait ReplicationAggregationSummary extends Summary {
-  val replicationSet: ReplicationSet
+  val replicaPath: String
```
kenoir: Value class would be slightly better than `String` (preventing certain modes of future misuse), and should incur no performance penalty. As per: https://docs.scala-lang.org/overviews/core/value-classes.html

```
class ReplicaPath(val underlying: String) extends AnyVal
```

replica_aggregator/src/main/scala/uk/ac/wellcome/platform/storage/replica_aggregator/services/ReplicaAggregator.scala
```diff
@@ -0,0 +1,67 @@
...
+      }
+
+      val startTime = Instant.now()
+
+      val replicaPath = result.storageLocation.location.path
+
+      versionedStore.upsert(replicaPath)(Set(result)) { existing =>
+        existing ++ Set(result)
+      } match {
+        // Only a single result is enough for now.
```
kenoir: 👍

PR #322, opened 2019-08-14
Disambiguate between ObjectLocation and the Prefix

alexwlchan:
It’s semantically cleaner if ObjectLocation only refers to a specific object that exists, and the Prefix refers to a directory or similar.

Unfortunately we have StorageLocation doing double-duty as both the source location (a tar.gz that exists in S3) and in the storage manifest.

I might finish this off later, see how I go.


kenoir:
Makes sense.

alexwlchan:
Replaced by #325.

PR #332, opened 2019-08-19
Refactor the replicator aggregator internals to distinguish between primary/secondary replicas

alexwlchan:


replica_aggregator/src/main/scala/uk/ac/wellcome/platform/storage/replica_aggregator/models/ReplicaResult.scala
```diff
@@ -3,29 +3,29 @@ package uk.ac.wellcome.platform.storage.replica_aggregator.models
...
         prefix = payload.bagRootLocation.asPrefix
       ),
       timestamp = Instant.now()
     )
 }
+
+
+
+// TODO: This needs moving somewhere!
+case class KnownReplicas(
```
kenoir: Shall we remove or comment this out for now?

PR #343, opened 2019-08-21
Use better error handling for update in new storage lib

kenoir:


replica_aggregator/src/main/scala/uk/ac/wellcome/platform/storage/replica_aggregator/services/ReplicaAggregatorWorker.scala
```diff
@@ -37,7 +34,7 @@ class ReplicaAggregatorWorker[IngestDestination, OutgoingDestination](
     ] {
 
   private sealed trait WorkerError
-  private case class AggregationFailure(e: Throwable) extends WorkerError
+  private case class AggregationFailure[E <: StorageError](e: E) extends WorkerError
```
alexwlchan: Why the type parameter? Will it ever be anything except `UpdateError`?

PR #351, opened 2019-08-21
Stop this bag unpacker test from being flaky

alexwlchan:
If it happened to create a valid S3 bucket name, the exception message wouldn't match, and the test would fail.

It’s just failed on master: https://travis-ci.org/wellcometrust/storage-service/jobs/574958614

### DIFF
--- a/bag_unpacker/src/test/scala/uk/ac/wellcome/platform/archive/bagunpacker/services/s3/S3UnpackerTest.scala
+++ b/bag_unpacker/src/test/scala/uk/ac/wellcome/platform/archive/bagunpacker/services/s3/S3UnpackerTest.scala
@@ -1,6 +1,7 @@
 package uk.ac.wellcome.platform.archive.bagunpacker.services.s3
 
 import com.amazonaws.services.s3.AmazonS3
+import com.amazonaws.services.s3.model.AmazonS3Exception
 import org.scalatest.Assertion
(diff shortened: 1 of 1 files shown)


bag_unpacker/src/test/scala/uk/ac/wellcome/platform/archive/bagunpacker/services/s3/S3UnpackerTest.scala
```diff
@@ -98,9 +101,9 @@ class S3UnpackerTest extends UnpackerTestCases[Bucket] with S3Fixtures {
 
           val underlyingError =
             ingestResult.asInstanceOf[IngestFailed[UnpackSummary]]
-          underlyingError.e shouldBe a[Throwable]
+          underlyingError.e shouldBe a[AmazonS3Exception]
           underlyingError.e.getMessage should startWith(
-            "The specified bucket is not valid"
+            "The specified bucket does not exist"
```
jamesgorrie: Should you update the test name to indicate we're looking for existence over failure?

alexwlchan: If you scroll up, you'll see the test name already says that:

```scala
it("fails if asked to write to a non-existent bucket") {
```

But we'd never noticed it was checking the wrong message!

PR #354, opened 2019-08-22
Lock the replicator around the destination, not the ingest ID

alexwlchan:
If you lock around the ingest ID, the Glacier and IA replications can't run simultaneously, and there's a 5-hour timeout on the queue. >.<

If you're writing to different destinations, it's fine to run replications concurrently. Desirable, even!

Discovered in #352.


alexwlchan:
Travis doesn't seem to have picked this up. 😕

PR #356, opened 2019-08-22
Expose inner error for IngestStoreUnexpectedError

kenoir:


kenoir:
> This looks fine, but it’s not obvious to me what (if anything) this changes. Where would we see an error where we wouldn't have seen one before?

The IngestStoreErrors, nest a throwable, they also extend a throwable - when this error is printed the inner error is hidden by the outer one.

`s"IngestStoreUnexpectedError: ${e.toString}"`

PR #359, opened 2019-08-23
Use the TCP logstash appender as UDP is hard to debug

kenoir:


kenoir:
Can't be sure that packets are not turning up at the logstash transit server because _mystery network thing_.

PR #360, opened 2019-08-28
A retryable error in the replica aggregator => retry operation

alexwlchan:
Tentative fix for https://github.com/wellcometrust/platform/issues/3860


replica_aggregator/src/main/scala/uk/ac/wellcome/platform/storage/replica_aggregator/services/ReplicaAggregatorWorker.scala
```diff
@@ -69,6 +69,17 @@ class ReplicaAggregatorWorker[IngestDestination, OutgoingDestination](
     val startTime = Instant.now()
 
     val ingestStep = getKnownReplicas(payload.replicaResult) match {
+      case Left(AggregationFailure(err: RetryableError)) =>
```
alexwlchan: This is wrong, I think? That’s an UpdateError; I want to catch an UpdateError that's wrapping a RetryableError.

PR #361, opened 2019-08-30
Unpacker changes for C&D

alexwlchan:


alexwlchan:
We need to close the connection when we're done.

PR #366, opened 2019-09-06
Slightly nicer errors when dealing with DDS

kenoir:
Differentiate between JSON decoding and request status errors and provide some output to help debug.

e.g.

```
AWS_PROFILE=storage ./migtool.sh simulate_goobi_calls --delay 0 --filter '3/0/0/0'
[
Unexpected error: <class 'Exception'>
Traceback (most recent call last):
  File "../src/migration_tools.py", line 319, in <module>
    fire.Fire(MigrationTool)
  File "/Users/k/Library/Python/3.6/lib/python/site-packages/fire/core.py", line 127, in Fire
    component_trace = _Fire(component, args, context, name)
  File "/Users/k/Library/Python/3.6/lib/python/site-packages/fire/core.py", line 366, in _Fire
    component, remaining_args)
  File "/Users/k/Library/Python/3.6/lib/python/site-packages/fire/core.py", line 542, in _CallCallable
    result = fn(*varargs, **kwargs)
  File "../src/migration_tools.py", line 36, in simulate_goobi_calls
    call_dds(delay, filter, total, after)
  File "../src/migration_tools.py", line 272, in call_dds
    raise Exception(message)
Exception: Got non-200 response from https://library-uat.wellcomelibrary.org/goobipdf/b18230003!
Status code: 503
Raw response:

[START]<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.01//EN""http://www.w3.org/TR/html4/strict.dtd">
<HTML><HEAD><TITLE>Service Unavailable</TITLE>
<META HTTP-EQUIV="Content-Type" Content="text/html; charset=us-ascii"></HEAD>
<BODY><h2>Service Unavailable</h2>
<hr><p>HTTP Error 503. The service is unavailable.</p>
</BODY></HTML>
[END]
```

### DIFF
--- a/bagger/src/migration_tools.py
+++ b/bagger/src/migration_tools.py
@@ -1,14 +1,17 @@
-import fire
-import aws
-import time
-import uuid
 import datetime
 import dateutil
-import settings
-import requests
 import json
-import storage_api
+import requests
+import sys
+import time
+import uuid
+
+import fire
+import aws
+
+import settings
 import dds
+import storage_api
 import status_table
 import migration_report
 from mets_filesource import bnumber_generator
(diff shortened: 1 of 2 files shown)


bagger/src/migration_tools.py
```diff
@@ -1,14 +1,17 @@
...
-import aws
-import time
-import uuid
 import datetime
 import dateutil
-import settings
-import requests
 import json
-import storage_api
+import requests
```
alexwlchan: minor: requests isn't standard library.

bagger/src/migration_tools.py
```diff
@@ -255,8 +258,31 @@ def call_dds(delay, filter, total, after):
...
+                raise Exception(message)
+
+            j = response.json()
+
+        except ValueError:
+            raise Exception(
+                f"Decoding JSON from  {url} failed for:\n {response.text}\n"
+            )
+
+        except:
```
alexwlchan: minor: Catch `except Exception` here, not a raw `except`.

The latter will include things like SystemExit and KeyboardInterrupt, which you don't usually want to catch.

PR #373, opened 2019-09-10
Add IngestId to StorageManifest

kenoir:
So we can identify to which ingest a manifest relates.

More specifically this will allow us to identify duplicate messages received by the register.

**Note:** This will require us to truncate what the storage service knows about already as it changes the internal model.

### DIFF
--- a/scripts/ss_report_stats.py
+++ b/scripts/ss_report_stats.py
@@ -69,7 +69,7 @@ for page in paginator.paginate(TableName="storage-ingests"):
             last_event = max(
                 event["createdDate"] for event in item["payload"]["events"]
             )
-        except KeyError:
+        except (KeyError, TypeError):
             last_event = item["payload"]["createdDate"]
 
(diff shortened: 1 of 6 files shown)


scripts/ss_report_stats.py
```diff
@@ -69,7 +69,7 @@ def draw_chart(data, colors=None):
             last_event = max(
                 event["createdDate"] for event in item["payload"]["events"]
             )
-        except KeyError:
+        except (KeyError, TypeError):
```
kenoir: Sneaky!

alexwlchan:
Tests? I think this needs at least two test cases:

* If you send the register the same space/externalIdentifier/version and ingest ID pair twice, it quietly ignores the second
* If you send the register the same space/externalIdentifier/version and two different ingest IDs, it errors out the second

kenoir:
@alexwlchan this PR is just adding the ingestId, not changing behaviour.

PR #380, opened 2019-09-30
Migration Reporter Mega Branch [DO NOT MERGE]

kenoir:
Pushed some of the Jupyter work into a script to isolate and share the DDS status updating.


scripts/status_reporter/status_store.py
```diff
@@ -0,0 +1,137 @@
+import sqlite3
+
+
+class StatusStore:
+    def _chunks(self, data, rows=10000):
```
alexwlchan: Minor: you could use the `@staticmethod` decorator here.

i.e.

```python
@staticmethod
def _chunks(data, rows=10000):
    for i in range(0, len(data), rows):
        yield data[i : i + rows]
```

Although this has nothing to do with this class; I'd be tempted to pull it out into a helper file.

scripts/status_reporter/aws_client.py
```diff
@@ -0,0 +1,30 @@
+import uuid
+
+import boto3
+
+
+class AwsClient:
+    def __init__(self, role_arn="arn:aws:iam::975596993436:role/storage-developer"):
```
alexwlchan: Really nice.

We use `attrs` in a couple of places for this sort of thing – sort of similar to case classes in Scala, which makes this a bit cleaner, but not strictly necessary.

scripts/status_reporter/aws_client.py
```diff
@@ -0,0 +1,30 @@
...
+        self.role_arn = role_arn
+
+    def _assumed_role_session(self):
+        session = boto3.session.Session()
+        sts_client = session.client("sts")
+
+        assumed_role_object = sts_client.assume_role(
+            RoleArn=self.role_arn, RoleSessionName=str(uuid.uuid1().hex)
```
alexwlchan: If we wanted to be more informative, the getpass module has a getuser() method that returns the name of the currently logged in user, so sessions could be tagged with our macOS usernames.

But I don't think I've ever looked at these session names, so not a big deal.

scripts/status_reporter/bnumbers.py
```diff
@@ -0,0 +1,54 @@
...
+
+            for obj in contents:
+                key = obj["Key"]
+                if key.startswith(prefix) and key.endswith(suffix):
+                    yield obj
+
+            try:
+                kwargs["ContinuationToken"] = resp["NextContinuationToken"]
+            except KeyError:
+                break
```
alexwlchan: This is oooolllllldddd code, use paginates instead: https://alexwlchan.net/2019/07/listing-s3-keys/

(Also the suffix thing isn’t something we need, we can ditch it)

scripts/status_reporter/bnumbers.py
```diff
@@ -0,0 +1,54 @@
...
+
+    def _get_matching_s3_keys(self, client, bucket, prefix="", suffix=""):
+        for obj in self._get_matching_s3_objects(client, bucket, prefix, suffix):
+            yield obj["Key"]
+
+    def bnumbers(self):
+        bucket = self.bucket
+        prefix = self.prefix
+
+        bnumber_pattern = re.compile(r"\A" + prefix + r"[0-9ax/]*/(b[0-9ax]{8}).xml\Z")
```
alexwlchan: What’s this pattern trying to match? I don’t recognise `\A` or `\Z` (are they like `^` and `$`?).

Prefer named groups to numeric groups (i.e. `(?P<bnumber>b[0-9ax]{8})`).

Minor: you probably want `\.xml` rather than a wildcard.

scripts/status_reporter/bnumbers.py
```diff
@@ -0,0 +1,54 @@
...
+
+        for key in self._get_matching_s3_keys(self.s3_client, self.bucket, self.prefix):
+            match = bnumber_pattern.match(key)
+            if bnumber_pattern.match(key):
```
alexwlchan: Worth a log if a key doesn't match the pattern, so we know we aren't missing some?

scripts/status_reporter/bnumbers.py
kenoir: This is duplicated from the bagger - i'm reluctant to case a different spell here.

scripts/status_reporter/__main__.py
```diff
@@ -0,0 +1,218 @@
...
+        if requested is "requested":
+            updated_status = dds.status(bnumber)
+            if assert_updated:
+                assert updated_status["status"] is "waiting"
+
+            return True
+        else:
+            return False
+    else:
+        return False
```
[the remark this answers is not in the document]
kenoir: Whether the ingest was requested successfully. 

This is pretty nasty (nesty) code though - so assigning to a variable with a descriptive name and returning that might improve things.

scripts/status_reporter/aws_client.py
kenoir: Should we be using `attrs` as standard practise in this situation?

kenoir:
Closing as not required.

PR #382, opened 2019-10-14
Adds LargeResponses trait

kenoir:
Allows routes providing responses in excess of a given byte length to store those responses in S3 and provide redirection to a signed S3 URL.

### DIFF
--- a/common/src/main/scala/uk/ac/wellcome/platform/archive/common/storage/LargeResponses.scala
+++ b/common/src/main/scala/uk/ac/wellcome/platform/archive/common/storage/LargeResponses.scala
@@ -1,0 +1,79 @@
+package uk.ac.wellcome.platform.archive.common.storage
+
+import java.net.URL
+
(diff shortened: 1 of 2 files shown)


common/src/main/scala/uk/ac/wellcome/platform/archive/common/storage/LargeResponses.scala
```diff
@@ -0,0 +1,79 @@
...
+  private val converter = StreamConverters.asInputStream()
+
+  implicit val materializer: ActorMaterializer
+
+  def wrapLargeResponses(route: Route): Route =
+    mapResponse(storeAndRedirect)(route)
+
+  private def storeAndRedirect(response: HttpResponse): HttpResponse = {
+    response.entity.contentLengthOption match {
+      case Some(length) if length > maximumResponseByteLength => {
```
alexwlchan: Maybe worth a debug log that we've gone down this path?

PR #384, opened 2019-10-14
Add existence check for S3Uploader

kenoir:

### DIFF
--- a/common/src/main/scala/uk/ac/wellcome/platform/archive/common/storage/services/S3ObjectExists.scala
+++ b/common/src/main/scala/uk/ac/wellcome/platform/archive/common/storage/services/S3ObjectExists.scala
@@ -1,0 +1,29 @@
+package uk.ac.wellcome.platform.archive.common.storage.services
+
+import com.amazonaws.services.s3.AmazonS3
+import uk.ac.wellcome.storage.{ObjectLocation, StorageError, StoreReadError}
+
+import scala.util.{Failure, Success, Try}
+
+trait ObjectExists {
+  def exists(objectLocation: ObjectLocation): Either[StorageError, Boolean]
+}
+
+class S3ObjectExists(s3Client: AmazonS3) extends ObjectExists {
+  override def exists(objectLocation: ObjectLocation): Either[StoreReadError, Boolean] = Try {
+    s3Client.doesObjectExist(
+      objectLocation.namespace, objectLocation.path
+    )
+  } match {
+    case Success(exists) => Right(exists)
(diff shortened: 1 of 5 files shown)


common/src/main/scala/uk/ac/wellcome/platform/archive/common/storage/services/S3ObjectExists.scala
```diff
@@ -0,0 +1,35 @@
...
+trait ObjectExists {
+  def exists(objectLocation: ObjectLocation): Either[StorageError, Boolean]
+}
+
+class S3ObjectExists(s3Client: AmazonS3) extends ObjectExists {
+  override def exists(
+    objectLocation: ObjectLocation
+  ): Either[StoreReadError, Boolean] =
+    Try {
+      s3Client.doesObjectExist(
```
alexwlchan: TIL: this method exists, so handy!

common/src/test/scala/uk/ac/wellcome/platform/archive/common/storage/services/S3UploaderTest.scala
```diff
@@ -35,6 +35,107 @@ class S3UploaderTest
...
+        .getLastModified()
+
+      getUrl(url) shouldBe content
+
+      Thread.sleep(1500)
+
+      val newUrl = uploader
+        .uploadAndGetURL(
+          location = objectLocation,
+          content = content,
```
alexwlchan: Ditto above. And assert that getting `url` now gets the new content!

alexwlchan:
And maybe an info-level log of which approach we took, so we can see if it's doing something daft?

common/src/test/scala/uk/ac/wellcome/platform/archive/common/storage/services/S3ObjectExistsTest.scala
```diff
@@ -0,0 +1,59 @@
...
+        }
+      }
+    }
+
+    describe("when there is an error retrieving from S3") {
+      it("returns a StorageError") {
+        val objectLocation =
+          ObjectLocation(randomAlphanumeric, randomAlphanumeric)
+
+        objectLocation.exists.left.value shouldBe a[StorageError]
```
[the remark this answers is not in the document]
alexwlchan: Damnit, GitHub/train WiFi swallowed my comment.

If this test is relying on the fact that `randomAlphanumeric` returns something that isn't a bucket name, be wary – sometimes it does! Use `createInvalidBucketName` instead.

PR #385, opened 2019-10-15
Store & redirect large responses

kenoir:

### DIFF
--- a/api/bags_api/src/main/scala/uk/ac/wellcome/platform/storage/bags/api/BagsApi.scala
+++ b/api/bags_api/src/main/scala/uk/ac/wellcome/platform/storage/bags/api/BagsApi.scala
@@ -29,15 +29,16 @@ import uk.ac.wellcome.platform.storage.bags.api.models.{
   ResultListEntry
 }
 import uk.ac.wellcome.storage.{NoVersionExistsError, ReadError}
-
+import uk.ac.wellcome.platform.archive.common.storage.LargeResponses
 import scala.concurrent.ExecutionContext
 import scala.util.matching.Regex
 import scala.util.{Failure, Success, Try}
 
-trait BagsApi extends Logging {
+trait BagsApi extends LargeResponses with Logging {
 
   import akka.http.scaladsl.server.Directives._
   import de.heikoseeberger.akkahttpcirce.FailFastCirceSupport._
+
 
   implicit val printer: Printer =
     Printer.noSpaces.copy(dropNullValues = true)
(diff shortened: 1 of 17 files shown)


api/bags_api/src/main/scala/uk/ac/wellcome/platform/storage/bags/api/BagsApi.scala
```diff
@@ -129,19 +130,14 @@ trait BagsApi extends Logging {
             )
 
           case Right(manifests) =>
-            val etagValue = manifests.map(_.idWithVersion).mkString("&")
-            val etag = ETag(etagValue)
```
alexwlchan: What happened to ethe tag here?

api/bags_api/src/main/scala/uk/ac/wellcome/platform/storage/bags/api/Main.scala
```diff
@@ -4,20 +4,31 @@ import java.net.URL
...
+import uk.ac.wellcome.storage.typesafe.S3Builder
 import uk.ac.wellcome.typesafe.WellcomeTypesafeApp
 import uk.ac.wellcome.typesafe.config.builders.AkkaBuilder
 
 import scala.concurrent.ExecutionContext
+import scala.concurrent.duration._
 
 object Main extends WellcomeTypesafeApp {
+  val defaultCacheDuration = 1 days
+  // 9MB = 1048576 Bytes * 9
```
alexwlchan: Why 9MB? Could we get a link to the relevant API Gateway docs?

api/bags_api/src/test/scala/uk/ac/wellcome/platform/storage/bags/api/BagsApiFeatureTest.scala
```diff
@@ -137,7 +137,7 @@ class BagsApiFeatureTest
               val header: ETag = response.header[ETag].get
               val etagValue = header.etag.value.replace("\"", "")
 
-              etagValue shouldBe storageManifest.idWithVersion
+              etagValue shouldBe s"${storageManifest.space}/${storageManifest.info.externalIdentifier}/${storageManifest.version}"
```
alexwlchan: 👍 much better

api/bags_api/src/main/scala/uk/ac/wellcome/platform/storage/bags/api/BagsApi.scala
kenoir: The tests flushed out that we are unlikely to get large multi version responses (as they will likely result from having _many_ versions), and if we do assembling the key like this will result in extremely long and possibly invalid keys.

PR #386, opened 2019-10-21
Add some code for downloading the contents of a bag

alexwlchan:
This isn't especially hard, but it is a little fiddly and something we'll probably want to do over and over again.

Closes https://github.com/wellcometrust/platform/issues/3949

### DIFF
--- a/python_client/src/wellcome_storage_service/downloader.py
+++ b/python_client/src/wellcome_storage_service/downloader.py
@@ -21,8 +21,7 @@ def download_bag(storage_manifest, out_dir):
 
     """
     all_files = (
-        storage_manifest["manifest"]["files"] +
-        storage_manifest["tagManifest"]["files"]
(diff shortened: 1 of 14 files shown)


python_client/src/wellcome_storage_service/downloader.py
```diff
@@ -0,0 +1,86 @@
...
+
+        mkdir_p(os.path.dirname(out_path))
+
+        with open(out_path, "wb") as file_obj:
+            self._download_fileobj(
+                location=location, manifest_file=manifest_file, file_obj=file_obj
+            )
+
+
+class S3InfrequentAccessProvider(AbstractProvider):
```
kenoir: Fancy!

alexwlchan:
This is now released as v1.3.0: https://pypi.org/project/wellcome-storage-service/1.3.0/

PR #395, opened 2019-10-30
Prevent dev deletions/writes in the prod storage service

alexwlchan:
We need to think about how we do deletions properly; I've started recording some thoughts in https://github.com/wellcometrust/platform/issues/4002. For now that's not a big concern, and we don't want to accidentally delete any of the digitised content.

Requires https://github.com/wellcometrust/terraform-modules/pull/171

Closes https://github.com/wellcometrust/platform/issues/3948

### DIFF
--- a/terraform/modules/critical/vhs_manifests.tf
+++ b/terraform/modules/critical/vhs_manifests.tf
@@ -1,0 +1,4 @@
+module "vhs_manifests" {
+  source = "git::https://github.com/wellcometrust/terraform-modules.git//vhs/modules/vhs?ref=v19.16.2"
+  name   = "${var.namespace}-manifests"
+}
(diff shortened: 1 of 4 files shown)


terraform/modules/critical/vhs_manifests.tf
```diff
@@ -1,4 +1,4 @@
 module "vhs_manifests" {
-  source = "git::https://github.com/wellcometrust/terraform-modules.git//vhs/modules/vhs?ref=v19.13.6"
+  source = "git::https://github.com/wellcometrust/terraform-modules.git//vhs/modules/vhs?ref=4660e69"
```
kenoir: this probably needs to be a fixed version -  we can do that in a future PR though

PR #397, opened 2019-10-30
Apply consistent naming to resources in our Terraform stack

alexwlchan:
Some prep work for https://github.com/wellcometrust/platform/issues/4006; the naming has got a bit out-of-date and I don't want to delete the wrong thing by accident!

* References to "access bucket" are now "primary replica"
* References to "archive bucket" are now "Glacier replica"
* The bag register is reading from the primary replica, and not (as it was previously, oops) the Ireland replica
* The wc-storage-ingests bucket is now wc-storage-unpacked-bags, which is a name that better describes how we're using it. It's also moved from `critical` to `stack`, because it only contains temporary data
* The storage-manifests VHS is named much more consistently in the Terraform state

Still to do:

* [x] Finish emptying and deleting the wc-storage-ingests bucket.


alexwlchan:
Looking at this some more, I realise that the better name is really “replica_glacier”, not “replica_ireland”. Will fix.

PR #401, opened 2019-11-05
Include tagmanifest-sha256.txt in the storage manifests created in the bag register

alexwlchan:
Code fix for https://github.com/wellcometrust/platform/issues/4012

This will ensure that all new bags we register get a `tagmanifest-sha256.txt` entry in the `tagManifest` field of the storage service manifest, where previously they didn't.

Backfilling old bags will be handled separately.

(Too many manifests!)


kenoir:
👏 Good catch.

PR #404, opened 2019-11-07
Compress a downloaded bag under a single directory, add modified dates

alexwlchan:
Closes https://github.com/wellcometrust/platform/issues/4007, closes https://github.com/wellcometrust/platform/issues/4008

I don't think we need to create directory entries explicitly; I think `tarfile.addfile()` does it implicitly. We'll find out when we deploy it!

For https://github.com/wellcometrust/platform/issues/4004

### DIFF
--- a/python_client/src/wellcome_storage_service/downloader.py
+++ b/python_client/src/wellcome_storage_service/downloader.py
@@ -74,8 +74,7 @@ def download_compressed_bag(storage_manifest, out_path):
 
             # Set the modified time to the creation dateb of the storage manifest
             creation_date = dt.datetime.strptime(
-                storage_manifest["createdDate"],
-                "%Y-%m-%dT%H:%M:%S.%fZ"
+                storage_manifest["createdDate"], "%Y-%m-%dT%H:%M:%S.%fZ"
             )
             tarinfo.mtime = time.mktime(creation_date.timetuple())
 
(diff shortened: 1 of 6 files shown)


python_client/src/wellcome_storage_service/downloader.py
```diff
@@ -57,15 +59,25 @@ def download_compressed_bag(storage_manifest, out_path):
...
             )
 
-            tarinfo = tarfile.TarInfo(name=manifest_file["name"])
+            tarinfo = tarfile.TarInfo(
+                name=os.path.join(ext_identifier, manifest_file["name"])
+            )
             tarinfo.size = manifest_file["size"]
 
+            # Set the modified time to the creation dateb of the storage manifest
+            creation_date = dt.datetime.strptime(
```
helenst: This discards the time zone info from the timestamp which results in offset timestamps on the tar info. Which I only noticed because Archivematica's  Django settings specify a US time zone. In python 3 we could use %z to parse the time zone but not in py 2. Can we force the datetime object to utc here if it's safe to assume all timestamps from the storage service will be UTC?

alexwlchan: We could, but these timestamps are already somewhat arbitrary – the storage manifest timestamp is when it's marked as done in the storage service, which can vary by hours depending on how much other stuff is in flight.

It would be nice to improve this, but I'm not too fussed about it.

PR #408, opened 2019-11-19
Use nginx defined in the storage service

alexwlchan:
The storage service was pulling in an nginx image from the catalogue account, so when the permissions on that ECR repo changed, we could no longer start tasks using that image.

The bags API fell over this morning (out-of-memory from requesting a really big bag), so this PR brings a copy of the nginx image into this repo, and uses this in the tasks.

### DIFF
--- a/nginx/Dockerfile
+++ b/nginx/Dockerfile
@@ -1,0 +1,6 @@
+FROM nginx:1.15.8-alpine
+
+COPY nginx.conf /etc/nginx/nginx.conf.template
+
+RUN apk add --update bash gettext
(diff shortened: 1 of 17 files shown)


nginx/Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM nginx:1.15.8-alpine
```
kenoir: I wonder if we could simple make the original public in docker hub and then build our own from that so we get consistency (same base for everyone) and resiliency (we are using a project specific ECR repo).

alexwlchan: We have https://github.com/wellcometrust/dockerfiles/tree/master/nginx, but it doesn't immediately work here and I don't have time to look at this right now.

This at least gets us more resiliency than before (when a change in the catalogue repo broke the storage service). Are you okay if I merge this as-is?

PR #410, opened 2019-11-25
Allow METS adapter to read from storage service

alicefuzier:


terraform/critical_prod/locals.tf
```diff
@@ -3,9 +3,11 @@ locals {
 
   goobi_task_role_arn         = "arn:aws:iam::299497370133:role/goobi_task_role"
   archivematica_task_role_arn = "arn:aws:iam::299497370133:role/am-prod-storage-service_task_role"
+  catalogue_pipeline_task_role-arn = "arn:aws:iam::760097843905:role/mets_adapter_task_role"
```
alexwlchan: 1. Could this have a better name? This is the task role ARN of the METS adapter, not the catalogue pipeline. Maybe `catalogue_mets_adapter_task_role_arn`?
2. Nit: `role_arn`, not `role-are`.

alicefuzier: The name was intentional. I don't think it should matter from the storage service which application of the catalogue stack is actually making the call. If we change it and have another app make the request or if we create an assumable role, they would still be `catalogue_pipeline_task_role_arn`, so I think this name is better. It keeps as much knowledge as possible of the catalogue pipeline out of storage.

But, if you feel strongly about, I'll change it.

alicefuzier:
This is applied

PR #411, opened 2019-11-25
Fix an aggregator bug that meant retryable errors were being ignored

alexwlchan:
This is caused by a subtle mismatch between the behaviour of the VersionedStore type class and the aggregator.

When the VersionedStore sees a retryable error in update() or upsert(), it wraps the underlying error as:

    new UpdateWriteError(err) with RetryableError

But the aggregator code was pattern matching on a different expression:

    case UpdateWriteError(err: RetryableError)

These aren't the same thing, and the mismatch meant these errors were still slipping through.  This should fix this source of flakiness in the replica aggregator.

See: https://github.com/wellcometrust/scala-storage/blob/95626ad8570a322421fcfd58e98bd50ed8016798/storage/src/main/scala/uk/ac/wellcome/storage/store/VersionedStore.scala#L41-L44

Closes https://github.com/wellcometrust/platform/issues/4074


replica_aggregator/src/main/scala/uk/ac/wellcome/platform/storage/replica_aggregator/services/ReplicaAggregatorWorker.scala
```diff
@@ -84,6 +88,18 @@ class ReplicaAggregatorWorker[IngestDestination, OutgoingDestination](
           e = err.e
         )
 
+      case Left(AggregationFailure(err: RetryableError)) =>
```
kenoir: 👏

PR #412, opened 2019-12-02
Don't allow slashes in the storage service

alexwlchan:
Supporting slashes adds a bunch of edge cases and weirdnesses that we'd need to check (e.g. around S3 paths and URL path components). If we had time to test it thoroughly, we could do it, but for now reject any ingest that tries to use "a/b" or "b/c" as a space or external ID.

---

Prompted by thinking about Archivematica, where the Calm IDs do contain slashes (but there's no way we want to pass those through).

The Archivematica side should never be sending them; to be on the safe side the storage service will reject them if it does.


alexwlchan:
@kenoir Perhaps I can make the comment clearer – the correct behaviour probably is escaping, but testing that the escaping works and doesn’t have any edge cases is at least a week of work. And thus not worth the time right now.

This doesn’t preclude us doing escaping in future, but it should catch silly bugs from the Archivematica side.

alexwlchan:
This is an improvement on the current behaviour, which is “handle it inconsistently and in a possibly broken way”.

PR #417, opened 2019-12-03
Improve validation of StorageSpace and ExternalIdentifier

alexwlchan:
If you try to create a:

* Storage space with a slash (e.g. `alfa/bravo`)
* An empty storage space `StorageSpace("")`
* An empty external identifier `ExternalIdentifier("")`

You hit these new runtime assertions. In practice, you should never hit these unless you're doing something very silly – but for https://github.com/wellcometrust/platform/issues/4088 in particular, we're going to have to assume spaces can't contain slashes, and it's better to encode that in the type system than assuming it won't happen.

This allows us to unambiguously distinguish the storage space/external identifier in a bag ID.

e.g. if you look at bag ID `alfa/bravo/charlie`, the space is `alfa` and the external ID is `bravo/Charlie`.

### DIFF
--- a/common/src/main/scala/uk/ac/wellcome/platform/archive/common/bagit/models/ExternalIdentifier.scala
+++ b/common/src/main/scala/uk/ac/wellcome/platform/archive/common/bagit/models/ExternalIdentifier.scala
@@ -35,7 +35,7 @@ object ExternalIdentifier {
     cursor.value.as[String].map(ExternalIdentifier(_))
 
   implicit def evidence: DynamoFormat[ExternalIdentifier] =
-    DynamoFormat.iso[ExternalIdentifier, String](
+    DynamoFormat.coercedXmap[ExternalIdentifier, String, IllegalArgumentException](
       ExternalIdentifier(_)
     )(
       _.underlying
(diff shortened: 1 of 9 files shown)


common/src/main/scala/uk/ac/wellcome/platform/archive/common/bagit/models/ExternalIdentifier.scala
```diff
@@ -3,19 +3,39 @@ package uk.ac.wellcome.platform.archive.common.bagit.models
...
   override def toString: String = underlying
+
+  require(!underlying.isEmpty, "External identifier cannot be empty")
+
+  // Normally we use case classes for immutable data, which provide these
+  // methods for us.
+  //
+  // We deliberately don't use case classes here so we skip automatic
+  // case class derivation for JSON encoding/DynamoDB in Scanamo,
+  // and force callers to intentionally import the implicits below.
```
alexwlchan: Previously we got this through AnyVal, which we've had to abandon because AnyVal doesn't let you do runtime checks.

alexwlchan:
Rationale: if we start allowing slashes in external identifiers, the assumption that storage space never contains a slash becomes pretty critical.

We'll enforce it at the ingests API stage – but hey, why not enforce it in the type system too. What are types for if not this? :)

alexwlchan:
(We could in theory support storage spaces with slashes too, by URL encoding it before putting it at the start of an S3 key – but there's much less motivation to do so.)

PR #418, opened 2019-12-04
Allow the storage service to scale to 0 when not running

alexwlchan:
Most of the time the storage service isn't doing anything, but we have Fargate instances running anyway. This is just burning money – allowing it to scale to 0 will save us ~$1200/month.

Applying this config showed up a bug in autoscaling setup – despite it being configured in Terraform, the notifier and ingests monitor both had broken autoscaling, so they'd never run anything except 1 instance.

Closes https://github.com/wellcometrust/platform/issues/4014. I did try writing a script to automate warming the pipeline, but let's see if we need it first.


kenoir:
Niiiice.

PR #427, opened 2019-12-16
Allow catalogue pipeline to subscribe to register_output topic

alicefuzier:
For https://github.com/wellcometrust/platform/issues/4045

### DIFF
--- a/terraform/modules/stack/iam_policy_document.tf
+++ b/terraform/modules/stack/iam_policy_document.tf
@@ -138,7 +138,7 @@ data "aws_iam_policy_document" "archivematica_ingests_get" {
 
 // This is the default policy that gets added to a topic when no policy is supplied
 // plus the statement that allows other principals to subscribe to this topic
-data "aws_iam_policy_document" "register_output_subscribe" {
+data "aws_iam_policy_document" "bag_register_output_subscribe" {
   policy_id = "__default_policy_ID"
 
   // default permissions copied from https://www.terraform.io/docs/providers/aws/r/sns_topic_policy.html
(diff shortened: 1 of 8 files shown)


terraform/modules/stack/iam_policy_document.tf
```diff
@@ -133,3 +135,58 @@ data "aws_iam_policy_document" "archivematica_ingests_get" {
...
+      "SNS:DeleteTopic",
+      "SNS:AddPermission",
+    ]
+
+    condition {
+      test     = "StringEquals"
+      variable = "AWS:SourceOwner"
+
+      values = [
+        "${data.aws_caller_identity.current.account_id}",
```
alexwlchan: nit: in terraform 0.12, we get to drop the quotes around things, and that's what we've been doing
```suggestion
        data.aws_caller_identity.current.account_id,
```

terraform/modules/stack/variables.tf
```diff
@@ -107,3 +107,7 @@ variable "archivematica_ingests_bucket" {
 
 variable "min_capacity" {}
 variable "max_capacity" {}
+
+variable "register_output_subscribe_principals" {
+  type = list
```
alexwlchan: We can type this stronger, to get better error reporting if we screw it up:

```suggestion
  type = list(string)
```

terraform/modules/stack/messaging.tf
```diff
@@ -315,6 +315,12 @@ module "bag_register_output_topic" {
   ]
 }
 
+resource "aws_sns_topic_policy" "register_topic_subscribe_policy" {
```
jamesgorrie: Worth calling it `bag_register_...` as I read it below and was trying to figure out what it was.

PR #428, opened 2020-01-13
Goobi script tasks need read access

kenoir:
Goobi has some more tasks that need read access to the storage bucket.

### DIFF
--- a/terraform/critical_staging/main.tf
+++ b/terraform/critical_staging/main.tf
@@ -1,7 +1,7 @@
 module "critical" {
   source = "../modules/critical"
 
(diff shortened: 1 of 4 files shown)


terraform/critical_staging/main.tf
```diff
@@ -1,16 +1,23 @@
 module "critical" {
   source = "../modules/critical"
 
-  namespace = "${local.namespace}-staging"
```
alexwlchan: Err… are you sure you want to do this? Won't this cause all the bucket/database names to collide?

PR #438, opened 2020-02-13
Try sharing the Travis cache between jobs

alexwlchan:
Travis uses the environment variable to determine the cache: if two jobs have different environment variables, they get different caches. See https://docs.travis-ci.com/user/caching/#caches-and-build-matrices

This means every job is having to rebuild the common lib and download dependencies.  This is very inefficient!  Let's try sharing the cache among all Scala jobs and see if that improves build times.

h/t Jamie for spotting this originally


alexwlchan:
Closing – this doesn't seem to make any noticeable effect on build times, but it does make it harder to see which job in the list has failed.

PR #440, opened 2020-02-13
Put a lot more detail in the README

alexwlchan:
The previous README was very light on detail.

This steals some words from my Stacks post (which was “why”, “what” and “how”), and expands the README to explain “what” and a bit of “why”. Plus linking to all the other things that have been written, so they can all be found in one place.

I’ll write something else about the architecture/concepts (e.g. what is an “ingest”), but I’ll do that in a separate PR.


alexwlchan:
TODO:

* Link to RFC
* Link to open issues
* Developer docs

README.md
```diff
@@ -2,6 +2,128 @@
...
+    This is typically an identifier from another system, which matches this bag to that record.
+    Examples: `b31497652`, `PP/CRI/A/2`.
+
+*   **Version:** an auto-incrementing version assigned by the storage service.
+    This tracks distinct versions of a (space, external identifier) pair.
+    Examples: `v1`, `v2`, `v3`.
+
+These three parts can be combined into a single string, which uniquely identifies a bag; for example `digitised/b31497652/v2`.
+This identifier is also the path to the root of the bag inside our storage buckets.
+
```
jamesgorrie: I'm a little unclear as to what parts of the ID are supplied vs calculated.

__Space__: Does the client define the space?
__Source ID__: Easy, supplied by client
 __Version__: Assuming this is controlled by the storage service?

jamesgorrie:
> Link to open issues

What for? Are you not asking to be updating docs every time you close an issue?

alexwlchan:
> What for?

The issues/known problems/whatever are associated with this repo and so useful to somebody looking at it, but they're kept in the platform repo.

> Are you not asking to be updating docs every time you close an issue?

Not if I link to the [storage-service] tag in our issue list. 😉

jamesgorrie:
Yea nice, I read it as linking to specific open issues in the readme, and then re-read it in the way you've explained...

README.md
alexwlchan: Added another line; make more sense now?

jamesgorrie: Nothing to do with the docs, but the space can be anything?

alexwlchan: The space doesn't allow slashes (and if you try to store a bag while passing a space that contains a slash, it gets rejected). Everything else is okay.

Not allowing slashes in the space/version means that if you have an identifier like

```
born-digital/PP/CRI/A/2/v1
```

you can unambiguously decide which bit is the space, external identifier and version.

(If slashes were allowed, you wouldn't be able to distinguish between a space `born-digital`, `born-digital/PP` or even `born-digital/PP/CRI`.)

And it has to be a valid S3 key, which I think is all Unicode but I guess one day we might find out otherwise? 🤷‍♀

alexwlchan: See https://github.com/wellcomecollection/storage-service/blob/master/common/src/main/scala/uk/ac/wellcome/platform/archive/common/storage/models/StorageSpace.scala (which I'll explain in the developer docs)

PR #441, opened 2020-02-17
Remove the `bin` directory

alexwlchan:
I wasn't using anything here, it's not very well-documented and we have better/alternative tooling for everything in here. Propose we ditch it all; we can rewrite anything we need fairly easily.


jamesgorrie:
There's broken tests, but assuming that's nothing to do with the PR but rather the Travis setup?

alexwlchan:
Yeah, the bag verifier has a known flaky test (https://github.com/wellcomecollection/platform/issues/3952).

PR #446, opened 2020-02-25
When we read a stream, hash it with all four BagIt checksum algorithms

alexwlchan:
This lays some early groundwork for https://github.com/wellcomecollection/platform/issues/4277, which I'd like to see done sooner rather than later

In particular, when we read a file out of S3, we create MD5/SHA1/SHA256/SHA512 checksums all in one go, rather than having to read it four times.

Right now we don't use the extra values anywhere, but later we can read the other manifests, and start comparing the checksums they contain from the values we get from S3.

### DIFF
--- a/common/src/main/scala/uk/ac/wellcome/platform/archive/common/storage/services/TagManifestFileFinder.scala
+++ b/common/src/main/scala/uk/ac/wellcome/platform/archive/common/storage/services/TagManifestFileFinder.scala
@@ -58,7 +58,8 @@ class TagManifestFileFinder[IS <: InputStream with HasLength](
         // so it happens after the entire bag has been verified.  To verify the bag,
         // we've already had to read the tagmanifest-sha256.txt file, so an error
         // here would be unlikely (but probably not impossible).
-        val checksum = Hasher.hash(is.identifiedT)
+        val checksum = Hasher
+          .hash(is.identifiedT)
           .get
           .getChecksumValue(algorithm)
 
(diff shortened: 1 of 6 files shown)


kenoir:
I like the direction here!

kenoir:
Pinging @jtweed to prioritise.

jtweed:
I'd put it below the born digital migration and requesting.

common/src/main/scala/uk/ac/wellcome/platform/archive/common/storage/services/TagManifestFileFinder.scala
```diff
@@ -52,10 +49,20 @@ class TagManifestFileFinder[IS <: InputStream with HasLength](
...
       case Right(is) =>
+        // TODO: we could handle the checksum more gracefully here, not '.get'
+        //
+        // This method is called in the bag register to create the storage manifest,
+        // so it happens after the entire bag has been verified.  To verify the bag,
+        // we've already had to read the tagmanifest-sha256.txt file, so an error
+        // here would be unlikely (but probably not impossible).
+        val checksum = Hasher
+          .hash(is.identifiedT)
+          .get
```
kenoir: Can we at least catch and re-throw with a more sensible error?

alexwlchan: Is good suggestion, will change.

PR #454, opened 2020-03-05
Use Fargate Spot for the storage service

alexwlchan:
This patch uses https://github.com/wellcomecollection/terraform-aws-ecs-service/pull/6 to use Fargate Spot in some of the storage service apps, specifically:

* the bag root finder
* the bag versioner
* the replica aggregator
* the bag register
* the notifier
* the ingests monitor

because all these tasks are SQS-backed and can be interrupted at short notice. The unpacker, verifier and replicator all stay as they are.

It also turns on spot for the APIs **in the staging service only**, on the basis that it's an easy cost win for something where we don't need 100% uptime.

Still needs to be rolled out to prod, which turns out to be a bit more delicate if we want to avoid downtime.

### DIFF
--- a/terraform/stack_prod/provider.tf
+++ b/terraform/stack_prod/provider.tf
@@ -4,6 +4,6 @@ provider "aws" {
   }
 
   region  = var.aws_region
-  version = "2.34.0"
+  version = "2.51.0"
 }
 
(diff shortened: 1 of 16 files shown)


terraform/stack_prod/provider.tf
```diff
@@ -4,6 +4,6 @@ provider "aws" {
   }
 
   region  = var.aws_region
-  version = "2.34.0"
+  version = "2.51.0"
```
jamesgorrie: You can just use `~> 2.0`. I've never found locking TF stuff necessary as they version really well.

alexwlchan:
@jamieparkinson 

> Pipeline next...? 😉

If I knew which bits of the pipeline would tolerate spot I would, but my knowledge is pretty out-of-date at this point. But it’s just dropping in a new module and some variables, how hard can it be? 😉

alexwlchan:
(Huh, I thought I'd merged this…)

PR #458, opened 2020-03-12
[Proof of concept] Run the storage service locally with docker-compose and localstack

alexwlchan:
The PR in its current state proves that it works, and also shows how much more there is to do.

- [x] Allow specifying the endpoint as environment variables for S3/DynamoDB (scala-storage)
- [x] Allow specifying the endpoint as environment variables for SNS/SQS (scala-messaging)
- [ ] Use the container that provides the ECS task metadata endpoint (https://github.com/awslabs/amazon-ecs-local-container-endpoints, h/t Robert)
- [ ] Get the remaining containers working
- [ ] Document the process
- [ ] Automate the bootstrapping of tables/queues
- [ ] How do you see what's inside the localstack environment?
- [ ] How do you upload bags?
- [ ] How would somebody who isn't a Wellcome dev get Docker images?

Closes https://github.com/wellcomecollection/platform/issues/4368


alexwlchan:
It turns out the flags for specifying AWS endpoints/access keys already exist, they're just several layers down: https://github.com/wellcomecollection/wellcome-typesafe-app/blob/master/src/main/scala/uk/ac/wellcome/typesafe/config/builders/AWSClientConfigBuilder.scala

alexwlchan:
I'm going to close this for now. Interesting experiment, but not something we've spent much time on since March and I think we're unlikely to do so any time soon.

PR #460, opened 2020-03-18
Add machinery to read more than one manifest file

alexwlchan:
This patch sets the stage to parse *all* the manifest files in a bag (not just the SHA256 manifest). Another stepping stone towards https://github.com/wellcomecollection/platform/issues/4277

In brief:

* The `CombinedManifest` trait can store the information from all the manifests in a bag, not just one. It then has two instances (terminology?) – `PayloadManifest` and `TagManifest`, so we can distinguish these at the type level.

   It stores data as `Map[BagPath, VerifiableChecksum]` rather than `List[BagFile]`, which has nicer semantics and reduces the amount of duplicated information.

* The `CombinedManifestParser` object can parse all the manifests in a bag, check their file lists are consistent, combine the information and return enough info to build a `CombinedManifest`.

  Plus it improves the validity checking we do on manifest files, and comes with shiny tests!

This patch drops in the new parser behind the scenes of the existing `BagManifest` model; if that works I need to go through and plumb in the new types.

### DIFF
--- a/common/src/main/scala/uk/ac/wellcome/platform/archive/common/bagit/models/CombinedManifestParser.scala
+++ b/common/src/main/scala/uk/ac/wellcome/platform/archive/common/bagit/models/CombinedManifestParser.scala
@@ -24,7 +24,7 @@ import scala.util.matching.Regex
   *     https://tools.ietf.org/html/rfc8493#section-2.2.1
   *
   */
-object ManifestFileParser {
+object CombinedManifestParser {
   def createFileLists(
     md5: Option[InputStream] = None,
     sha1: Option[InputStream] = None,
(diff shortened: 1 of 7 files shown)


common/src/main/scala/uk/ac/wellcome/platform/archive/common/bagit/models/CombinedManifestParser.scala
```diff
@@ -0,0 +1,166 @@
...
+      .mapValues { _.size }
+      .filter { case (_, count) => count > 1 }
+      .map { case (bagPath, _) => bagPath }
+      .toList
+
+    // NOTE: If a manifest contains the same file path twice but with two different
+    // checksums, this will only keep the last one.  Strictly speaking, it's an
+    // error for a manifest file to refer to the same file twice, and we should
+    // fail parsing here if that occurs.
+    println(parsedLines)
```
alexwlchan: This println() needs to go.

alexwlchan:
Closing as outdated; will be replaced by another PR.

PR #462, opened 2020-03-19
Remove unused code for writing a fetch.txt

alexwlchan:
This code was written when we thought the storage service might create its own fetch files for the replicas; we later decided not to do this, so we can remove these `write()` methods.

The storage service should only ever be reading BagIt metadata files, never creating its own.

An easy win spun out of https://github.com/wellcomecollection/storage-service/pull/461


alexwlchan:
Closing in favour of #461, since Robert reviewed that as a whole.

PR #463, opened 2020-03-23
Improve the error messages and testing of the bag-info.txt parser

alexwlchan:
Closes https://github.com/wellcomecollection/platform/issues/3764

I started looking at improving these error messages, and the more I tugged on the thread the more I realised there were some other code improvements I could make. Highlights:

* The bag info parsing code is now in its own object and tested using the same method as the calling code (previously the tests were using a special method that nobody else was using)

* Make bag-info.txt parsing better:

    - Handle duplicate labels/values (which are allowed by the BagIt spec but which may upset our storage service)
    - If the External-Identifier is an illegal value (e.g. `/a`), say so rather than throwing an uncaught exception
    - Don't allow empty labels; do allow empty lines
    - More human-readable error messages throughout, rather than a list of labels that failed for some reason

  We've lost some of the cats stuff which would summarise all the errors for you; I think that's an acceptable tradeoff. The user-facing message is usually truncated to ~140 characters, so you wouldn't see more than one of the new, more detailed error messages.

* Remove some unused classes and traits


common/src/test/scala/uk/ac/wellcome/platform/archive/common/bagit/models/BagInfoTest.scala
```diff
@@ -1,148 +1,99 @@
...
-    BagInfo.parseBagInfo(t, IOUtils.toInputStream(bagInfoString, "UTF-8")) shouldBe Left(
-      InvalidBagInfo(
-        t,
-        List("External-Identifier", "Payload-Oxum", "Bagging-Date")
-      )
-    )
-  }
-
-}
+//package uk.ac.wellcome.platform.archive.common.bagit.models
```
kenoir: Think you might have meant to remove this.

alexwlchan: 🤦‍♀ 

Oops, will fix, thanks.

PR #468, opened 2020-03-26
Experiment: initial attempt at a bag indexer

alexwlchan:
This is an initial experiment, but I want to go through it with @kenoir when he’s feeling better.

For https://github.com/wellcomecollection/platform/issues/4303

### DIFF
--- a/bag_indexer/src/main/scala/uk/ac/wellcome/platform/archive/bag_indexer/elasticsearch/ElasticsearchIndexCreator.scala
+++ b/bag_indexer/src/main/scala/uk/ac/wellcome/platform/archive/bag_indexer/elasticsearch/ElasticsearchIndexCreator.scala
@@ -1,7 +1,10 @@
 package uk.ac.wellcome.platform.archive.bag_indexer.elasticsearch
 
 import com.sksamuel.elastic4s.ElasticDsl._
-import com.sksamuel.elastic4s.requests.indexes.{CreateIndexResponse, PutMappingResponse}
+import com.sksamuel.elastic4s.requests.indexes.{
+  CreateIndexResponse,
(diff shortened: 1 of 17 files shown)


bag_indexer/src/main/scala/uk/ac/wellcome/platform/archive/bag_indexer/elasticsearch/ElasticsearchIndexCreator.scala
```diff
@@ -0,0 +1,81 @@
...
+          // This forces all our records to be indexed into a single shard,
+          // which should avoid this problem.
+          //
+          // If/when we index more records, we should revisit this.
+          //
+          .shards(1)
+      }
+      .flatMap { response: Response[CreateIndexResponse] =>
+        if (response.isError) {
+          if (response.error.`type` == "resource_already_exists_exception" || response.error.`type` == "index_already_exists_exception") {
```
kenoir, line 39: Can we un-nest/use pattern matching more liberally here?

kenoir:
I'm guessing this is missing the actual Worker implementation as we don't have a notion of where these are coming from yet.

alexwlchan:
> I'm guessing this is missing the actual Worker implementation as we don't have an exact idea of where the things that feed this service are coming from yet.

Actually for the bags service it already exists – the bag register has been sending notifications of newly registered bags for a while now.

It’s not quite the version we’ll use – the Elastic stuff went in separately, and I need to rejig the repo layout a bit – but yeah, expect to see this code in a revised form soon!

alexwlchan:
Oh, and I wanted to play with the indexing because I think that’s the interesting part of the bag indexer.

PR #469, opened 2020-04-06
Add the common Elasticsearch code to the storage service

alexwlchan:
A first step towards indexing stuff in Elasticsearch.

Spun out of https://github.com/wellcomecollection/storage-service/pull/468


alexwlchan:
> We should document where the indexer will fit!

We should, but first we should discuss it! Expect to see a proposal in a bit.

PR #474, opened 2020-04-07
The ingests monitor sends the full body of updated ingests

alexwlchan:
Follows https://github.com/wellcomecollection/storage-service/pull/473. First step of https://github.com/wellcomecollection/platform/issues/4417.

This patch modifies the ingests service to send the complete ingest to SNS every time it gets an update. It also includes the Terraform wiring, and a queue so we can see if it’s doing so successfully.

### DIFF
--- a/ingests/src/main/scala/uk/ac/wellcome/platform/archive/ingests/services/IngestsWorker.scala
+++ b/ingests/src/main/scala/uk/ac/wellcome/platform/archive/ingests/services/IngestsWorker.scala
@@ -78,7 +78,7 @@ class IngestsWorker[CallbackDestination, UpdatedIngestsDestination](
     val updatedIngestResult = updatedIngestsMessageSender.sendT(ingest)
 
     (callbackResult, updatedIngestResult) match {
-      case (Success(_), Success(_))   => Success(())
+      case (Success(_), Success(_)) => Success(())
 
       case (Failure(callbackErr), Success(_)) =>
         warn(s"Failed to send the callback notification: $callbackErr")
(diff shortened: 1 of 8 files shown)


ingests/src/main/scala/uk/ac/wellcome/platform/archive/ingests/services/IngestsWorker.scala
```diff
@@ -63,5 +65,27 @@ class IngestsWorker[CallbackDestination](
...
+  //
+  // These two messages are independent, and so should succeed/fail independently.
+  //
+  private def sendOngoingMessages(ingest: Ingest): Try[Unit] = {
+    val callbackResult = callbackNotificationService.sendNotification(ingest)
+    val updatedIngestResult = updatedIngestsMessageSender.sendT(ingest)
+
+    (callbackResult, updatedIngestResult) match {
+      case (Success(_), Success(_))   => Success(())
+      case _ =>
```
kenoir, line 82: We can be specific about the failure using pattern matching, so we probably should be.

ingests/src/main/scala/uk/ac/wellcome/platform/archive/ingests/services/IngestsWorker.scala
```diff
@@ -63,5 +65,27 @@ class IngestsWorker[CallbackDestination](
...
+        warn(s"One or both of the ongoing messages failed to send correctly:")
```
jamesgorrie, line 83: not `error`? Or do we `warn` before we `error` as the `Throwable` will `error` further up?

PR #476, opened 2020-04-08
Ensure ingest updates are ordered

alexwlchan:
We want we only put the newest version of an Ingest in Elasticsearch.

The Ingest model doesn't have a version, but it does have a modified date (the latest event it received), or the created date as a fallback. Use that field as the version in Elasticsearch to ensure updates are ordered correctly.

Useful links:

* The Ingest model. https://github.com/wellcomecollection/storage-service/blob/master/common/src/main/scala/uk/ac/wellcome/platform/archive/common/ingests/models/Ingest.scala
* The Indexer class in the catalogue that this code is based on: https://github.com/wellcomecollection/catalogue/blob/master/pipeline/ingestor/ingestor_common/src/main/scala/uk/ac/wellcome/platform/ingestor/common/Indexer.scala

Note: in the catalogue we index the internal model and then convert it to a display model in the Catalogue API; here we're indexing the display models directly, because we expect the primary interface for this data to be Kibana (at least, for now).

Part of https://github.com/wellcomecollection/platform/issues/4417

### DIFF
--- a/indexer/ingests_indexer/src/main/scala/uk/ac/wellcome/platform/archive/indexer/ingests/IngestIndexer.scala
+++ b/indexer/ingests_indexer/src/main/scala/uk/ac/wellcome/platform/archive/indexer/ingests/IngestIndexer.scala
@@ -27,7 +27,7 @@ class IngestIndexer(val client: ElasticClient, val index: Index)(
 
   override protected def version(ingest: Ingest): Long =
     ingest.lastModifiedDate match {
-      case Some(modifiedDate) => modifiedDate.getEpochSecond
-      case None               => ingest.createdDate.getEpochSecond
+      case Some(modifiedDate) => modifiedDate.toEpochMilli
+      case None               => ingest.createdDate.toEpochMilli
     }
 }
--- a/indexer/ingests_indexer/src/main/scala/uk/ac/wellcome/platform/archive/indexer/ingests/IngestIndexer.scala
+++ b/indexer/ingests_indexer/src/main/scala/uk/ac/wellcome/platform/archive/indexer/ingests/IngestIndexer.scala
@@ -24,4 +24,10 @@ class IngestIndexer(val client: ElasticClient, val index: Index)(
       ingest = ingest,
       contextUrl = new URL("http://localhost:9200")
     )
+
+  override protected def version(ingest: Ingest): Long =
+    ingest.lastModifiedDate match {
+      case Some(modifiedDate) => modifiedDate.getEpochSecond
+      case None               => ingest.createdDate.getEpochSecond
+    }
 }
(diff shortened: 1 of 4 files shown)


indexer/ingests_indexer/src/main/scala/uk/ac/wellcome/platform/archive/indexer/ingests/IngestIndexer.scala
```diff
@@ -24,4 +24,10 @@ class IngestIndexer(val client: ElasticClient, val index: Index)(
       ingest = ingest,
       contextUrl = new URL("http://localhost:9200")
     )
+
+  override protected def version(ingest: Ingest): Long =
+    ingest.lastModifiedDate match {
```
jamesgorrie, line 29: Like this a lot for versioning!

indexer/ingests_indexer/src/main/scala/uk/ac/wellcome/platform/archive/indexer/ingests/IngestIndexer.scala
```diff
@@ -24,4 +24,10 @@ class IngestIndexer(val client: ElasticClient, val index: Index)(
...
+      case Some(modifiedDate) => modifiedDate.getEpochSecond
+      case None               => ingest.createdDate.getEpochSecond
+    }
 }
```
jamesgorrie, line 33: Maybe using `toEpochMilli`?

indexer/ingests_indexer/src/main/scala/uk/ac/wellcome/platform/archive/indexer/ingests/IngestIndexer.scala
```diff
@@ -24,4 +24,10 @@ class IngestIndexer(val client: ElasticClient, val index: Index)(
       ingest = ingest,
       contextUrl = new URL("http://localhost:9200")
     )
+
+  override protected def version(ingest: Ingest): Long =
+    ingest.lastModifiedDate match {
+      case Some(modifiedDate) => modifiedDate.getEpochSecond
```
jamesgorrie, line 30: Maybe using `toEpochMilli`? A lot can happen in a second.

indexer/ingests_indexer/src/test/scala/uk/ac/wellcome/platform/archive/indexer/ingests/IngestIndexerTest.scala
```diff
@@ -99,4 +101,137 @@ class IngestIndexerTest
...
+  describe("orders updates correctly") {
+    describe("when both ingests have a modified date") {
+      val ingestId = createIngestID
+
+      val olderIngest = createIngestWith(
+        id = ingestId,
+        events = Seq(
+          IngestEvent(description = "event 1", createdDate = Instant.ofEpochSecond(101))
+        ),
+        createdDate = Instant.ofEpochSecond(1)
```
jamesgorrie, line 114: Not sure if you want to update the tests too - shouldn't make a difference, but might be more self-documenting.

PR #478, opened 2020-04-08
Add a worker for the ingests indexer

alexwlchan:
This allows the ingests indexer to, well, actually index things. Read off an SQS queue and log; there's no onward message here.

### DIFF
--- a/indexer/ingests_indexer/src/main/resources/application.conf
+++ b/indexer/ingests_indexer/src/main/resources/application.conf
@@ -5,4 +5,4 @@ es.port=${?es_port}
 es.username=${?es_username}
 es.password=${?es_password}
 es.protocol=${?es_protocol}
-es.ingests.indexPrefix=${?es_ingests_index_prefix}
+es.ingests.index-prefix=${?es_ingests_index_prefix}
--- a/indexer/ingests_indexer/src/main/resources/application.conf
+++ b/indexer/ingests_indexer/src/main/resources/application.conf
@@ -1,0 +1,8 @@
+aws.metrics.namespace=${?metrics_namespace}
+aws.sqs.queue.url=${?queue_url}
+es.host=${?es_host}
+es.port=${?es_port}
+es.username=${?es_username}
+es.password=${?es_password}
+es.protocol=${?es_protocol}
+es.ingests.indexPrefix=${?es_ingests_index_prefix}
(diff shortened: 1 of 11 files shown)


alexwlchan:
Once this is merged and built by Travis, I can start wiring it up to Terraform. Exciting!

indexer/ingests_indexer/src/test/scala/uk/ac/wellcome/platform/archive/indexer/ingests/IngestsIndexerWorkerTest.scala
```diff
@@ -0,0 +1,67 @@
...
+    val ingest = createIngest
+
+    val badMapping = properties(
+      Seq(textField("name"))
+    )
+
+    withLocalElasticsearchIndex(badMapping) { index =>
+      // We need to wait a bit before trying to index the document, otherwise the
+      // test is flaky in CI.
+      Thread.sleep(1000)
```
jamieparkinson, line 55: Might it be worth moving this into the `withLocalElasticsearchIndex` fixture itself? Or, better, replacing it with some kind of `eventually { indexIsCreated }` in that fixture instead?

alexwlchan: Yeah, it's already there, this fix turned out to be a completely wrong guess at why CI was failing.

indexer/ingests_indexer/src/main/resources/application.conf
```diff
@@ -0,0 +1,8 @@
+aws.metrics.namespace=${?metrics_namespace}
+aws.sqs.queue.url=${?queue_url}
+es.host=${?es_host}
+es.port=${?es_port}
+es.username=${?es_username}
+es.password=${?es_password}
+es.protocol=${?es_protocol}
+es.ingests.indexPrefix=${?es_ingests_index_prefix}
```
jamieparkinson, line 8: this is the most spectacularly miniscule nit which you are free to ignore, but as a point of trivia Lighbend recommend using `kebab-case` for HOCON variable names https://github.com/lightbend/config/blob/master/HOCON.md#hyphen-separated-vs-camelcase

PR #481, opened 2020-04-20
Add a script for patching a bag in the storage service

alexwlchan:
This does the work of:

* Building the initial fetch file
* Updating the manifest
* Uploading the bag to S3
* Triggering an ingest

I'm gonna have to do this a bunch of times for the Preservica verification, so putting it here for visibility. Not sure we actually want it in the repo proper.


alexwlchan:
Closing as not-wanted, but we can find it again if we need it.

PR #486, opened 2020-04-28
Add a display model for standalone files

alexwlchan:
We need this to index files in Elasticsearch – we should make sure we're happy with this model before we write the code to actually index it.

Part of https://github.com/wellcomecollection/platform/issues/4410

### DIFF
--- a/display/src/main/scala/uk/ac/wellcome/platform/archive/display/files/DisplayStandaloneFile.scala
+++ b/display/src/main/scala/uk/ac/wellcome/platform/archive/display/files/DisplayStandaloneFile.scala
@@ -1,0 +1,57 @@
+package uk.ac.wellcome.platform.archive.display.files
+
+import io.circe.generic.extras.JsonKey
+import uk.ac.wellcome.platform.archive.common.storage.models.{StorageManifest, StorageManifestFile}
+
+// We index individual files into their own index in Elasticsearch.
+//
+// For example, somebody might want to query for all the JP2 files created in
+// the "digitised" space in a given month.
+//
+// We don't want to ingest just the DisplayFile model used in the bags API,
+// because that doesn't have any context about the space of external identifier
+// of this particular bag.
+
+case class DisplayAssociatedBag(
+  space: String,
+  externalIdentifier: String,
+  version: String,
+  createdDate: String,
+  @JsonKey("type") ontologyType: String = "AssociatedBag"
+)
+
+case object DisplayAssociatedBag {
+  def apply(storageManifest: StorageManifest): DisplayAssociatedBag =
+    DisplayAssociatedBag(
+      space = storageManifest.space.underlying,
+      externalIdentifier = storageManifest.info.externalIdentifier.underlying,
+      version = storageManifest.version.toString,
+      createdDate = storageManifest.createdDate.toString
+    )
+}
+
+case class DisplayStandaloneFile(
+  checksum: String,
+  name: String,
+  path: String,
+  size: Long,
+  bag: DisplayAssociatedBag,
+  @JsonKey("type") ontologyType: String = "File"
+)
+
+case object DisplayStandaloneFile {
+  def apply(file: StorageManifestFile, storageManifest: StorageManifest): DisplayStandaloneFile = {
+    require(
+      storageManifest.manifest.files.contains(file),
+      s"File ${file.name} is not part of storage manifest ${storageManifest.idWithVersion}"
+    )
+
+    DisplayStandaloneFile(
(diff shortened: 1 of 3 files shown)


display/src/main/scala/uk/ac/wellcome/platform/archive/display/files/DisplayStandaloneFile.scala
```diff
@@ -0,0 +1,57 @@
...
+// For example, somebody might want to query for all the JP2 files created in
+// the "digitised" space in a given month.
+//
+// We don't want to ingest just the DisplayFile model used in the bags API,
+// because that doesn't have any context about the space of external identifier
+// of this particular bag.
+
+case class DisplayAssociatedBag(
+  space: String,
+  externalIdentifier: String,
```
jamesgorrie, line 17: What is this?

alexwlchan: Can you be more specific? Do you mean the case class "DisplayAssociatedBag", or "externalIdentifier", or something else?

display/src/main/scala/uk/ac/wellcome/platform/archive/display/files/DisplayStandaloneFile.scala
```diff
@@ -0,0 +1,57 @@
...
+// We index individual files into their own index in Elasticsearch.
+//
+// For example, somebody might want to query for all the JP2 files created in
+// the "digitised" space in a given month.
+//
+// We don't want to ingest just the DisplayFile model used in the bags API,
+// because that doesn't have any context about the space of external identifier
+// of this particular bag.
+
+case class DisplayAssociatedBag(
```
jamesgorrie, line 15: Is it called an `AssociatedBag`  as it's a link to the bag that contains the file?

alexwlchan: Pretty much, yeah. Alternative names welcome!

display/src/main/scala/uk/ac/wellcome/platform/archive/display/files/DisplayStandaloneFile.scala
jamesgorrie: Apologies. `externalIdentifier`?

alexwlchan: It's the identifier in something like Sierra, CALM or whatever, supplied by the user when they initiate an upload. e.g. `b12345678`

See https://github.com/wellcomecollection/storage-service#identifiers for context

display/src/main/scala/uk/ac/wellcome/platform/archive/display/files/DisplayStandaloneFile.scala
jamesgorrie: Yea, we've got the same thing on the Merger, using images as an example.

We use `MergedImage` and `UnmergedImage`, which maybe isn't clearly described.

Thinking about it in that way (the thing, and the link to the thing), could you just have `Bag` and BagLink` or maybe `BagRel`?

display/src/main/scala/uk/ac/wellcome/platform/archive/display/files/DisplayStandaloneFile.scala
jamesgorrie: You've explained this to me, thanks for doing it again!

display/src/main/scala/uk/ac/wellcome/platform/archive/display/files/DisplayStandaloneFile.scala
alexwlchan: `Bag` is already used in a bunch of other places – I wonder if `ParentBag` would work? Because there is a sort of child–parent relationship between file and bag.

jamesgorrie: Why not just link? Sorry that comment wasn't very clear, I was suggesting that `Bag` is the [current `Bag`](https://github.com/wellcomecollection/storage-service/blob/master/common/src/main/scala/uk/ac/wellcome/platform/archive/common/bagit/models/Bag.scala) and this is a `Link` to that `Bag`.

So maybe `BagLink` implying relationship, but not implying the type (hierarchical).

alexwlchan:
Moving to draft while we continue to discuss what this looks like.

alexwlchan:
Closing as not wanted.

PR #487, opened 2020-04-28
Add an S3 bucket policy that blocks Put/Delete except for the replicator

alexwlchan:
This adds another level of protection to the objects in the critical storage buckets. Previously creating a new "S3 super-access" user would allow you to futz with the bucket, but now you need to create such a user *and* unhook this policy first.

This is a test ingest in staging to verify I haven't broken anything: https://wellcome-ingest-inspector.glitch.me/ingests/94745dbe-d74b-4f61-8f24-b0d44d3eec1f

### DIFF
--- a/terraform/critical_prod/main.tf
+++ b/terraform/critical_prod/main.tf
@@ -21,7 +21,8 @@ module "critical" {
     "arn:aws:iam::653428163053:user/echo-fs",
   ]
 
-  replica_glacier_read_principals = []
+  replicator_primary_task_role_arn = "arn:aws:iam::975596993436:role/storage-prod-bag-replicator_primary_task_role"
+  replicator_glacier_task_role_arn = "arn:aws:iam::975596993436:role/storage-prod-bag-replicator_glacier_task_role"
 
   # This gives us another layer of protection for the S3 buckets.
   #
--- a/terraform/modules/critical/s3_replica_glacier.tf
+++ b/terraform/modules/critical/s3_replica_glacier.tf
@@ -16,18 +16,23 @@ resource "aws_s3_bucket" "replica_glacier" {
   }
 }
 
-resource "aws_s3_bucket_policy" "replica_glacier_read" {
-  count = length(var.replica_glacier_read_principals) == 0 ? 0 : 1
-
+resource "aws_s3_bucket_policy" "replica_glacier_access" {
   bucket = aws_s3_bucket.replica_glacier.id
-  policy = data.aws_iam_policy_document.replica_glacier_readonly.json
+  policy = data.aws_iam_policy_document.replica_glacier_access.json
 }
 
-data "aws_iam_policy_document" "replica_glacier_readonly" {
+data "aws_iam_policy_document" "replica_glacier_access" {
+  # By default, we don't allow anybody to write or delete objects in the bucket.
+  #
+  # We only allow write permissions to certain, pre-approved IAM roles.  This is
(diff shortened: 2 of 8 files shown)


terraform/critical_prod/main.tf
```diff
@@ -21,7 +21,8 @@ module "critical" {
     "arn:aws:iam::653428163053:user/echo-fs",
   ]
 
-  replica_glacier_read_principals = []
+  replicator_primary_task_role_arn = "arn:aws:iam::975596993436:role/storage-prod-bag-replicator_primary_task_role"
```
jamesgorrie, line 24: Where are these created? Could we pull them from that state?

alexwlchan: We could, but you’d create circular references between the two state files.

Stack depends on critical (to get the bucket/database) names, but then critical would depend on stack (to get these names).

terraform/modules/critical/s3_replica_glacier.tf
```diff
@@ -16,18 +16,44 @@ resource "aws_s3_bucket" "replica_glacier" {
...
+  policy = data.aws_iam_policy_document.replica_glacier_access.json
 }
 
-data "aws_iam_policy_document" "replica_glacier_readonly" {
+data "aws_iam_policy_document" "replica_glacier_access" {
+  # By default, we don't allow anybody to write or delete objects in the bucket.
+  #
+  # We only allow write permissions to certain, pre-approved IAM roles.  This is
+  # enforced at the bucket level, so even creating a role with "S3 full access" will
+  # not allow you to modify bucket objects.
```
jamesgorrie, line 29: As reading from Glacier, especially in large data sizes, can be very very expensive, should we be restricting that too?

alexwlchan: The potential cost of pulling the entire cold copy out of storage is actually quite manageable.

This bucket is cycling all its objects to Glacier Deep Archive, which doesn't have expedited retrieval. The cost of retrieving the entire archive would be annoying but perfectly manageable:

$0.021 / GB * ~60 TB = ~$1300

https://aws.amazon.com/s3/pricing/

I don't think we'd want to do it very often, but it's not on the same tier of "never ever ever ever ever do this" as deleting or modifying the archive in-place.

terraform/critical_prod/main.tf
jamesgorrie: Maybe a note?

terraform/modules/critical/s3_replica_glacier.tf
```diff
@@ -16,18 +16,44 @@ resource "aws_s3_bucket" "replica_glacier" {
...
+    }
+  }
+
+  statement {
+    effect = "Allow"
+
     actions = [
-      "s3:List*",
-      "s3:Get*",
+      "s3:PutObject*",
```
kenoir, line 56: I think this causes issues as the replicator also needs to read.

terraform/modules/critical/s3_replica_primary.tf
```diff
@@ -16,12 +16,63 @@ resource "aws_s3_bucket" "replica_primary" {
...
+        "*",
+      ]
+    }
+  }
+
+  statement {
+    effect = "Allow"
+
+    actions = [
+      "s3:PutObject*",
```
kenoir, line 57: I think this causes issues as the replicator also needs to read (as above).

alexwlchan:
Closing as not wanted.

PR #490, opened 2020-04-29
Ingest Tracker Skeleton app

kenoir:
Part of https://github.com/wellcomecollection/platform/issues/4419


ingests_tracker/src/main/resources/application.conf
```diff
@@ -0,0 +1 @@
+aws.dynamo.tableName=${?archive_ingest_table_name}
```
alexwlchan, line 1: What we're using elsewhere:

```suggestion
aws.dynamo.tableName=${?ingests_table_name}
```

PR #494, opened 2020-04-29
Update to the newest version of all our libraries

alexwlchan:
I want to patch something in scala-storage…

but that brings in new versions of scalatest and other libraries…

so I tried to update everything else…

and now I'm in dependency hell.

Currently stuck on deprecation errors coming out of Circe.


alexwlchan:
Yeah, I'm hoping we don't have too many more "move the entire world" PRs like this.

alexwlchan:
merging as this is passing, and should only have minor conflicts with the PATCH pull request

PR #502, opened 2020-05-07
Attempt to get the ingests worker feature tests passing

alexwlchan:
This is also mixed with an attempt to pass around ActorSystem implicitly.

It passes locally.


alexwlchan:
Closing as not needed.

PR #504, opened 2020-05-11
Add infra to deploy new ingests service

kenoir:
For https://github.com/wellcomecollection/platform/issues/4429


terraform/modules/service/ingest/main.tf
```diff
@@ -0,0 +1,85 @@
...
+  service_name = var.service_name
+  cluster_arn  = var.cluster_arn
+
+  container_definitions = [
+    module.nginx_container.container_definition,
+    module.external_api_container.container_definition,
+    module.worker_container.container_definition
+  ]
+
+  desired_task_count = var.desired_task_count
```
alexwlchan, line 13: I can't find it now, but I'm sure we used to do something like this for the task counts in the API:

```suggestion
  desired_task_count = max(3, var.desired_task_count)
```

with a link to the bit of API Gateway documentation that said "you should run an API in every region".

terraform/modules/stack/iam_role_policy.tf
```diff
@@ -18,41 +18,41 @@ resource "aws_iam_role_policy" "bag_register_metrics" {
...
 }
 
 resource "aws_iam_role_policy" "ingests_metrics" {
-  role   = module.ingests.task_role_name
+  role   = module.ingest_service.task_role_name
   policy = data.aws_iam_policy_document.cloudwatch_putmetrics.json
 }
 
-# ingests_api
+# ingests_service
```
alexwlchan, line 47: Delete this comment, move it up to the next heading.

terraform/modules/stack/main.tf
```diff
@@ -1,7 +1,120 @@
...
+
+  security_group_ids = [
+    aws_security_group.interservice.id,
+    aws_security_group.service_egress.id
+  ]
+
+  cluster_arn = aws_ecs_cluster.cluster.arn
+
+  external_api_container_image = local.image_ids["ingests_api"]
+  internal_api_container_image = local.image_ids["ingests_tracker"]
```
alexwlchan, line 16: Use the API/tracker names? Let's be consistent.

kenoir: I've removed all the name references and am using this format everywhere as it felt redundant to have a name and a value in the map.

PR #509, opened 2020-05-14
Add a getIngest() method to the ingest tracker API

alexwlchan:
Part of modifying the external-facing ingests API to use the ingests tracker (https://github.com/wellcomecollection/platform/issues/4419).

I tweaked the test cases to make them more generic, and had to rename a few of the existing errors to distinguish between and update/get error.


ingests_tracker/src/test/scala/uk/ac/wellcome/platform/storage/ingests_tracker/client/IngestTrackerClientTest.scala
```diff
@@ -2,94 +2,178 @@ package uk.ac.wellcome.platform.storage.ingests_tracker.client
...
+          whenReady(future.failed) {
+            _ shouldBe a[Throwable]
+          }
+        }
+      }
     }
   }
 }
+
+class AkkaIngestTrackerClientTest
```
kenoir, line 168: 👍

PR #511, opened 2020-05-14
Use the internal IngestTracker API in the external-facing ingests API

alexwlchan:
The next step of https://github.com/wellcomecollection/platform/issues/4419

This wires up the ingests API code to talk to the internal tracker API, although it'll need new config and to be deployed as part of the new task definition for it to work.

We may be able to merge the `ingest_common` lib into the `ingest_tracker` app, but I'll leave that for a separate patch.


api/ingests_api/src/main/scala/uk/ac/wellcome/platform/storage/ingests/api/IngestsApi.scala
```diff
@@ -2,24 +2,43 @@ package uk.ac.wellcome.platform.storage.ingests.api
...
+    onComplete(future) {
+      case Success(resp) => resp
+      case Failure(_) =>
+        complete(
+          StatusCodes.InternalServerError -> InternalServerErrorResponse(
+            contextURL,
+            statusCode = StatusCodes.InternalServerError
+          )
+        )
+    }
```
alexwlchan, line 43: Based on https://github.com/wellcomecollection/catalogue/blob/2c3ebea20121ab7b649d4eb964b6f090ec9fe108/api/api/src/main/scala/uk/ac/wellcome/platform/api/rest/CustomDirectives.scala#L67-L76

api/ingests_api/src/main/scala/uk/ac/wellcome/platform/storage/ingests/api/responses/CreateIngest.scala
```diff
@@ -50,44 +52,68 @@ trait CreateIngest extends ResponseBase with Logging {
...
+  ): Future[Route] = {
+    val ingest = requestDisplayIngest.toIngest
+
+    val creationResult = for {
+      trackerResult <- ingestTrackerClient.createIngest(ingest)
+      _ <- trackerResult match {
+        case Right(_) => Future.fromTry { unpackerMessageSender.sendT(SourceLocationPayload(ingest)) }
+        case Left(_) => Future.successful(())
+      }
+    } yield trackerResult
```
alexwlchan, line 80: If the ingest gets to the database but not to the unpacker, we treat it as a failed creation, which seems fine to me – the user should try to create it again, or nothing will be stored.

This matches the previous behaviour.

api/ingests_api/src/main/scala/uk/ac/wellcome/platform/storage/ingests/api/responses/CreateIngest.scala
```diff
@@ -50,44 +54,73 @@ trait CreateIngest extends ResponseBase with Logging {
...
+              )
+            )
+          }
+
+        // This will capture both a Conflict and an Internal error from the ingest
+        // tracker.  There's nothing a user can do about either of them, so return
+        // them both as 500 errors.
+        case Left(_) =>
           complete(
-            StatusCodes.Created -> ResponseDisplayIngest(
```
kenoir, line 72: Should we pass along the conflict?

alexwlchan: I did consider it, but I think not.

You get a conflict if you try to create an Ingest and there's already an Ingest with that ID.

The IDs are assigned randomly when we convert a request ingest (from a user) to an internal ingest. Nothing a user does should affect whether you get a conflict here, so I figured better to map them both to 500s and have it show up on our dashboards.

PR #512, opened 2020-05-14
Delete the old ingests app and ingests_common lib

alexwlchan:
As part of https://github.com/wellcomecollection/platform/issues/4419, we can consolidate ingests_common into the new ingest_tracker app, and do away with the old ingests app. Codebase is cleaner, and we get back two builds in Travis!


alexwlchan:
This has passed in Travis, but GitHub Hooks are being sloooooow.

PR #514, opened 2020-05-15
Allow managing app clients in Terraform

alexwlchan:
Closes https://github.com/wellcomecollection/platform/issues/4519

I haven't ported all the existing ones because cycling the credentials is a bit tedious, but it makes it a lot easier to create new ones.


terraform/app_clients/secrets.tf
```diff
@@ -0,0 +1,25 @@
+locals {
+  description = "Storage service credentials; managed in https://github.com/wellcomecollection/storage-service/tree/master/terraform/app_clients"
+}
+
+# Store the Catalogue client credentials in KMS for the METS adapter
```
kenoir, line 5: Should these go in the platform account so they can be extracted through SSM easily?

alexwlchan: 1. They’re already in the platform account – it owns the Cognito User Pool, so that’s the Terraform provider I’m using.

2. The METS adapter already pulls from these SSM parameters, but now they’re managed automatically rather than somebody copying them across by hand.

PR #515, opened 2020-05-15
IngestsTrackerApi should be a class

kenoir:
This will hopefully simplify creating IngestsTrackerApi a bit and remove the possibility of "scala.UninitializedFieldError: Uninitialized field" error.


kenoir:
For https://github.com/wellcomecollection/platform/issues/4429

PR #522, opened 2020-05-18
Parameterise Indexer classes

kenoir:
Make it easier to create new IndexerWorkers.


indexer/common/src/test/scala/uk/ac/wellcome/platform/archive/indexer/IndexerWorkerTestCases.scala
```diff
@@ -0,0 +1,73 @@
...
+  NonDeterministicFailure,
+  Successful
+}
+import uk.ac.wellcome.platform.archive.common.generators.IngestGenerators
+import uk.ac.wellcome.platform.archive.indexer.elasticsearch.Indexer
+import uk.ac.wellcome.platform.archive.indexer.fixtures.IndexerFixtures
+
+abstract class IndexerWorkerTestCases[T, IndexedT](
+  implicit val decoder: Decoder[T]
+) extends AnyFunSpec
```
alexwlchan, line 20: Making this a class is a nice way to get the automatic decoder. 👍

indexer/common/src/test/scala/uk/ac/wellcome/platform/archive/indexer/IndexerWorkerTestCases.scala
```diff
@@ -0,0 +1,73 @@
...
+
+      val storedT =
+        getT[Json](index, id)
+          .as[Map[String, Json]]
+          .right
+          .value
+
+      val storedTId = storedT("id").asString.get
+
+      storedTId shouldBe id
```
alexwlchan, line 55: Suggestion: Could we instead check that IndexedT has been stored correctly?

This "id" stuff is a holdover from when we were indexing a slightly warped display model with the context removed, so there was no convenient case-class deserialiser.

indexer/common/src/test/scala/uk/ac/wellcome/platform/archive/indexer/fixtures/IndexerFixtures.scala
```diff
@@ -1,41 +1,38 @@
...
-    index: Index
-  )(testWith: TestWith[IngestsIndexerWorker, R]): R = {
-    val ingestIndexer = new IngestIndexer(
-      client = elasticClient,
-      index = index
-    )
-
+  def withIndexerWorker[R](
+    index: Index,
+    indexer: Index => Indexer[T, IndexedT],
```
alexwlchan, line 25: Could you require this be defined as a method on the class, rather than passing it around every time? Would it ever change for a given type T?

PR #526, opened 2020-05-19
Bag Indexer infrastructure

kenoir:

### DIFF
--- a/terraform/modules/stack/main.tf
+++ b/terraform/modules/stack/main.tf
@@ -81,6 +81,44 @@ module "ingests_indexer" {
   service_discovery_namespace_id = local.service_discovery_namespace_id
 }
 
+# bag_indexer
+
+module "bag_indexer" {
+  source = "../service/worker"
+
+  container_image = local.image_ids["bag_indexer"]
+
+  cluster_name = aws_ecs_cluster.cluster.name
+  cluster_arn  = aws_ecs_cluster.cluster.arn
+
+  subnets      = var.private_subnets
+  service_name = "${var.namespace}-bag-indexer"
+
+  environment = {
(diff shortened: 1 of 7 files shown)


terraform/modules/stack/main.tf
```diff
@@ -81,6 +81,44 @@ module "ingests_indexer" {
...
+  secrets = var.bag_indexer_secrets
+
+  security_group_ids = [
+    aws_security_group.service_egress.id,
+    aws_security_group.interservice.id
+  ]
+
+  # We run the indexer all the time to updates appear in the reporting cluster
+  # almost as soon as they're available in the API, rather than waiting for the
+  # indexer to spin up/down every time.
```
alexwlchan, line 113: Do we need bags to have the same availability as ingests?

terraform/stack_prod/main.tf
```diff
@@ -60,9 +60,19 @@ module "stack_prod" {
...
+    es_protocol = "prod/bag_indexer/es_protocol"
+    es_username = "prod/bag_indexer/es_username"
+    es_password = "prod/bag_indexer/es_password"
+  }
+
   workflow_bucket_name = local.workflow_bucket_name
 
   archivematica_ingests_bucket             = data.terraform_remote_state.archivematica_infra.outputs.ingests_bucket_arn
-  bag_register_output_subscribe_principals = ["${local.catalogue_pipeline_account_principal}"]
+  bag_register_output_subscribe_principals = [local.catalogue_pipeline_account_principal]
```
alexwlchan, line 76: Nice catch.

terraform/modules/stack/main.tf
kenoir: hmm, probably not

PR #529, opened 2020-05-20
Adds a load method to indexer worker

kenoir:


kenoir:
This does not fix the issue with trying to use a StorageManifest, but provides for a step where a StorageManifest representation can be loaded.

PR #531, opened 2020-05-21
Add a Lambda that applies tags to new objects written to storage buckets

alexwlchan:
Closes https://github.com/wellcomecollection/platform/issues/4488

I'm 98% sure that https://github.com/wellcomecollection/platform/issues/4535 will turn out to be false.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,8 +24,6 @@ jobs:
     - env: SBT_PROJECT=ingests_api
     - env: SBT_PROJECT=notifier
 
-    - env: TASK=s3_object_tagger-test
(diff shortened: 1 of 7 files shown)


.travis.yml
```diff
@@ -24,8 +24,6 @@ jobs:
     - env: SBT_PROJECT=ingests_api
     - env: SBT_PROJECT=notifier
 
-    - env: TASK=s3_object_tagger-test
```
alexwlchan, line 27: Something is busted in the Lambda test image, and I don't want to break Travis right now.

PR #533, opened 2020-05-22
Add a stub for the bag tracker app

alexwlchan:
This includes a minimal API and classes, but no actual implementations – those will come in separate PRs.

First step towards https://github.com/wellcomecollection/platform/issues/4534

### DIFF
--- a/bag_tracker/src/main/scala/uk/ac/wellcome/platform/archive/bag_tracker/BagTrackerApi.scala
+++ b/bag_tracker/src/main/scala/uk/ac/wellcome/platform/archive/bag_tracker/BagTrackerApi.scala
@@ -1,0 +1,65 @@
+package uk.ac.wellcome.platform.archive.bag_tracker
+
+import akka.actor.ActorSystem
+import akka.http.scaladsl.Http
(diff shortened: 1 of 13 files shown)


bag_tracker/src/main/scala/uk/ac/wellcome/platform/archive/bag_tracker/BagTrackerApi.scala
```diff
@@ -0,0 +1,50 @@
...
+        post {
+          entity(as[StorageManifest]) { manifest =>
+            println(manifest)
+            println("create the manifest!")
+            complete(StatusCodes.Created)
+          }
+        },
+        pathSuffix("versions" /) {
+          path(Segment / Remaining) { (space, remaining) =>
+            println(s"space = $space")
```
kenoir, line 34: does this need a GET?

alexwlchan: Yes, it does. Oops. Coming in the next pull request, so I'm not going to fix it here.

PR #540, opened 2020-05-22
Move the bags API to the top level of the repo

alexwlchan:
Tiny repo cleanup.


kenoir:
Maybe all the bag prefix things remaining should be in a "bags" folder together?

PR #543, opened 2020-05-26
Modify the bag register to use the bag tracker

alexwlchan:
Another part of https://github.com/wellcomecollection/platform/issues/4534

### DIFF
--- a/bag_register/src/main/scala/uk/ac/wellcome/platform/archive/bag_register/services/BagRegisterWorker.scala
+++ b/bag_register/src/main/scala/uk/ac/wellcome/platform/archive/bag_register/services/BagRegisterWorker.scala
@@ -4,6 +4,7 @@ import akka.actor.ActorSystem
 import io.circe.Decoder
 import software.amazon.awssdk.services.sqs.SqsAsyncClient
 import uk.ac.wellcome.messaging.sqsworker.alpakka.AlpakkaSQSWorkerConfig
+import uk.ac.wellcome.messaging.worker.models.Result
 import uk.ac.wellcome.messaging.worker.monitoring.metrics.MetricsMonitoringClient
 import uk.ac.wellcome.platform.archive.bag_register.models.RegistrationSummary
 import uk.ac.wellcome.platform.archive.common.KnownReplicasPayload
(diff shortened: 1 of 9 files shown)


bag_register/src/main/scala/uk/ac/wellcome/platform/archive/bag_register/services/BagRegisterWorker.scala
```diff
@@ -44,15 +54,23 @@ class BagRegisterWorker[IngestDestination, OutgoingDestination](
...
 
+      _ <- Future.fromTry {
+        outgoingPublisher.sendIfSuccessful(
+          result = registrationSummary,
+          outgoing = payload
+        )
+      }
     } yield registrationSummary
+
+  override def processMessage(
```
kenoir, line 72: Why aren't we using this?

alexwlchan: I’ll drop in a comment.

alexwlchan: See comment in https://github.com/wellcomecollection/storage-service/pull/543/commits/c7c725aa9e5029bda8185d8a94037bf7c5695936

kenoir: Perhaps this is a situation where we should turn a future into a try by waiting for it rather than circumvent `processMessage` - we're waiting for an internal REST API to return, maybe a hard time limit here makes sense.

alexwlchan: Would certainly simplify some of the internal APIs if we could Try instead of Future. I wonder if that should be pushed into the tracker client? We don't really use Future anywhere else; it's bits of Akka that are leaking.

kenoir: I think there should be an explicit "amount of time to wait for the tracker api to return somewhere" at your discretion where that it.

alexwlchan: Probably the tracker client? Say, the tracker API has to return within 20 seconds or we give up?

PR #554, opened 2020-06-01
Update ES mappings

kenoir:
For https://github.com/wellcomecollection/platform/issues/4557


indexer/bag_indexer/src/main/scala/uk/ac/wellcome/platform/archive/indexer/bags/BagsIndexConfig.scala
```diff
@@ -2,41 +2,49 @@ package uk.ac.wellcome.platform.archive.indexer.bags
...
 
-  private val payloadStatsFields: Seq[FieldDefinition] =
-    Seq(
-      objectField("fileSuffixTally").fields(suffixTallyFields),
-      intField("fileCount"),
-      longField("fileSize")
-    )
+  private val infoFields: Seq[FieldDefinition] = Seq(
+    keywordField("externalIdentifier"),
+    keywordField("payloadOxum"),
```
alexwlchan, line 22: It suddenly strikes me that Payload-Oxum contains both file count and file size as a single value, albeit a little harder to parse. Maybe we don't need that value again?

(Possibly we store a structured representation of the Payload-Oxum?)

PR #559, opened 2020-06-02
Correct permissions for api/indexer

kenoir:
The ~api &~ indexer does not require write access to the VHS.


terraform/modules/stack/iam_role_policy.tf
```diff
@@ -14,7 +14,7 @@ resource "aws_iam_role_policy" "bag_register_metrics" {
 
 resource "aws_iam_role_policy" "bags_api_vhs_manifests_readonly" {
   role   = module.bags_api.task_role_name
-  policy = var.vhs_manifests_readwrite_policy
+  policy = var.vhs_manifests_readonly_policy
```
alexwlchan, line 17: Nuh uh, the API task definition includes the bag tracker which does need read-write access.

PR #560, opened 2020-06-02
Remove the BagVerifierConfig case class

alexwlchan:
It's not used anywhere in the codebase.


indexer/reindex_bags_backlog.py
```diff
@@ -210,6 +210,7 @@ def _query(ids):
 
     return flat_list
 
+
```
kenoir, line 213: ⚪

PR #562, opened 2020-06-03
The object tagger adds new tags, rather than replacing old ones

alexwlchan:
A first step towards https://github.com/wellcomecollection/platform/issues/4562

* If the object already had tags, we append the Content-Type tag to the existing tags, rather than replacing them all
* If the object already has a Content-Type tag with a different value, throw an error rather than overwriting the existing value

### DIFF
--- a/s3_object_tagger/src/s3_object_tagger.py
+++ b/s3_object_tagger/src/s3_object_tagger.py
@@ -31,9 +31,9 @@ def get_objects(event):
         yield (s3_obj["bucket"]["name"], s3_obj["object"]["key"])
 
 
-def apply_tags(s3_client, *, bucket, key, tags):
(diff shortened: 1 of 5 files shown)


alexwlchan:
> Can we imagine a situation where this might be an issue (perhaps if the verifier also write tags)?

Hmm.

The replicator will write the initial object, and then this tag gets applied.

At some point after that the verifier will come along and add its own tag, and it will do the whole "is there a tag yet, y/n" dance. Depending on the delay between replicator and verifier, it might get a bit tight (and we've been bitten by S3 object consistency before). Hmm.

alexwlchan:
Closing as WONTFIX, this has exposed a consistency issue that we'll handle elsewhere.

PR #566, opened 2020-06-03
Use the new notifications sent by the bag register in the indexer

alexwlchan:
Follows #565, for https://github.com/wellcomecollection/platform/issues/4558

### DIFF
--- a/terraform/modules/stack/messaging.tf
+++ b/terraform/modules/stack/messaging.tf
@@ -400,29 +400,6 @@ module "registered_bag_notifications_queue" {
 
 # bag indexer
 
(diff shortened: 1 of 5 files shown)


terraform/modules/stack/messaging.tf
```diff
@@ -398,6 +398,29 @@ module "registered_bag_notifications_queue" {
   dlq_alarm_arn = var.dlq_alarm_arn
 }
 
+# bag indexer
+
+module "bag_indexer_input_queue" {
```
kenoir, line 403: I think this will orphan the other queue this used to use? Merge master & remove it?

PR #567, opened 2020-06-03
Clarify bag_indexer messaging

kenoir:
Use a dedicated queue for the bag_indexer, subscribed to the reindex script and bag_register ouput topics.

### DIFF
--- a/terraform/modules/stack/messaging.tf
+++ b/terraform/modules/stack/messaging.tf
@@ -411,7 +411,7 @@ module "bag_reindexer_output_topic" {
 module "bag_indexer_input_queue" {
   source = "../queue"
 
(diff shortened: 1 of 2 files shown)


terraform/modules/stack/messaging.tf
```diff
@@ -407,3 +407,19 @@ module "bag_reindexer_output_topic" {
 
   role_names = []
 }
+
+module "bag_indexer_input_queue" {
+  source = "../queue"
+
+  name = "${var.namespace}_bag_register_output"
```
alexwlchan, line 414:
```suggestion
  name = "${var.namespace}_bag_indexer_input"
```

PR #568, opened 2020-06-04
Allow index settings and pass nested doc default

kenoir:
The files attribute on bag docs can have up to ~970000 nested docs!

### DIFF
--- a/indexer/bag_indexer/src/main/scala/uk/ac/wellcome/platform/archive/indexer/bags/Main.scala
+++ b/indexer/bag_indexer/src/main/scala/uk/ac/wellcome/platform/archive/indexer/bags/Main.scala
@@ -46,7 +46,7 @@ object Main extends WellcomeTypesafeApp {
     val indexCreator = new ElasticsearchIndexCreator(
       elasticClient = elasticClient
     )
-    
+
     indexCreator.create(
       index = index,
       mappingDefinition = BagsIndexConfig.mapping,
--- a/indexer/bag_indexer/src/main/scala/uk/ac/wellcome/platform/archive/indexer/bags/Main.scala
+++ b/indexer/bag_indexer/src/main/scala/uk/ac/wellcome/platform/archive/indexer/bags/Main.scala
@@ -47,21 +47,14 @@ object Main extends WellcomeTypesafeApp {
       elasticClient = elasticClient
     )
 
+    val maxExpectedFilesOnABag = 1000000
+
     indexCreator.create(
       index = index,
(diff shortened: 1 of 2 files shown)


indexer/bag_indexer/src/main/scala/uk/ac/wellcome/platform/archive/indexer/bags/Main.scala
```diff
@@ -49,7 +49,11 @@ object Main extends WellcomeTypesafeApp {
 
     indexCreator.create(
       index = index,
-      mappingDefinition = BagsIndexConfig.mapping
+      mappingDefinition = BagsIndexConfig.mapping,
+      settings = Map(
+        // The largest number of files on a bag is ~ 970,000
```
alexwlchan, line 54: Let's put some more clues for a future reader in this comment:

```suggestion
        // The largest number of files on a bag is ~ 970,000 (see b19974760, aka
        // Chemist and Druggist)
        //
        // If we use the default limit, we get errors from Elasticsearch:
        //
        //      [illegal_argument_exception] The length of [files.path] field of
        //      [digitised/b19974760] doc of [storage_bags] index has exceeded [1000000] -
        //      maximum allowed to be analyzed for highlighting. This maximum can be
        //      set by changing the [index.highlight.max_analyzed_offset] index level setting.
        //      For large texts, indexing with offsets or term vectors is recommended!
        //
```

kenoir: that's actually the wrong error you've quoted now I look at it

alexwlchan: Well what's the right error? :)

indexer/bag_indexer/src/main/scala/uk/ac/wellcome/platform/archive/indexer/bags/Main.scala
```diff
@@ -47,9 +47,16 @@ object Main extends WellcomeTypesafeApp {
       elasticClient = elasticClient
     )
 
+    val maxExpectedFilesOnABag = 1000000
```
alexwlchan, line 50: Why are you defining this constant if you don't use it below?

kenoir: FFS typo from addressing the last review comment.

PR #572, opened 2020-06-05
Update reindex bag script to new format

kenoir:
Following bag indexer input format change.


indexer/reindex_bags_backlog.py
```diff
@@ -67,31 +67,15 @@ def _assumed_role_session(role_arn):
     session = _assumed_role_session(role_arn)
     return session.client(service_name)
 
-
 # The bag indexer only cares about space, externalIdentifier & version
```
alexwlchan, line 70: You can ditch this comment now, no?

PR #574, opened 2020-06-05
Allow republishing in reindex script

kenoir:
Adds an option to republish missing bags.

Also removes the unused, incorrectly named  `publish_all.sh`script.


indexer/reindex_bags_backlog.py
```diff
@@ -243,20 +243,35 @@ def publish(env, dry_run, role_arn):
...
     dynamodb_client = create_client("dynamodb", role_arn)
     elastic_client = create_elastic_client(role_arn, ES_SECRETS)
+    sns_client = create_client("sns", role_arn)
 
     bags_to_publish = get_latest_bags(dynamodb_client, config["table_name"])
     bags_to_confirm = [key for (key, value) in bags_to_publish.items()]
 
     not_indexed = confirm_indexed(elastic_client, bags_to_confirm, config["es_index"])
-    print(f"NOT INDEXED: {not_indexed}")
+    if len(not_indexed) > 0:
```
alexwlchan, line 263: nit: in Python, collections are implicitly True if non-empty, implicitly False if not. It would be more idiomatic to write:

```suggestion
    if not_indexed:
```

PR #581, opened 2020-06-08
Tag objects with their checksum values in the bag verifier

alexwlchan:
For https://github.com/wellcomecollection/platform/issues/4562

This is way too big to review in its current form – I’ll break it up into smaller parts. This code is too important to slide through with minimal review.

### DIFF
--- a/bag_verifier/src/main/scala/uk/ac/wellcome/platform/archive/bagverifier/fixity/FixityChecker.scala
+++ b/bag_verifier/src/main/scala/uk/ac/wellcome/platform/archive/bagverifier/fixity/FixityChecker.scala
@@ -48,8 +48,6 @@ trait FixityChecker extends Logging {
     val result: Either[FileFixityError, FileFixityCorrect] = for {
       objectLocation <- getObjectLocation(expectedFileFixity)
 
-      inputStream <- getInputStream(objectLocation, expectedFileFixity)
-
       _ <- verifySize(
         expectedFileFixity = expectedFileFixity,
         objectLocation = objectLocation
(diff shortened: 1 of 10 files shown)


bag_verifier/src/main/scala/uk/ac/wellcome/platform/archive/bagverifier/fixity/FixityChecker.scala
```diff
@@ -50,85 +120,175 @@ trait FixityChecker extends Logging {
...
-          e = FailedChecksumCreation(algorithm, e)
+          algorithm = algorithm,
+          size = size
         )
 
-      // Checksum does not match that provided
-      case Success(checksum) =>
-        if (checksum != expectedFileFixity.checksum)
-          FileFixityMismatch(
+      case Some(storedValue) if storedValue == fixityTagValue => {
```
kenoir, line 190: 👍

bag_verifier/src/main/scala/uk/ac/wellcome/platform/archive/bagverifier/fixity/FixityChecker.scala
```diff
@@ -50,85 +120,175 @@ trait FixityChecker extends Logging {
...
+    } yield result
+
+  private def writeChecksumTag(
+    expectedFileFixity: ExpectedFileFixity,
+    objectLocation: ObjectLocation,
+    fixityTags: Map[String, String]): Either[FileFixityCouldNotWriteTag, Unit] =
+      Try {
+        debug(s"Adding tags $fixityTags to $objectLocation")
+        tags
+          .update(objectLocation) { existingTags =>
```
kenoir: I think this is only a weak guarantee though as the tags are not "locked" during this checking operation. I think you probably already know that & I would agree this check is worth doing in any case.

PR #583, opened 2020-06-09
Change SizeFinder to return an Either, not a Try

alexwlchan:
Another piece building on #581, part of wellcomecollection/platform#4562

This gives us the nicer types from scala-storage, so we can distinguish between "does not exist" and "unexpected error".  Having this distinction allows us to use these classes in the bag verifier, when we want to look up the size of an object that's been cycled to Glacier.

### DIFF
--- a/common/src/main/scala/uk/ac/wellcome/platform/archive/common/storage/services/SizeFinder.scala
+++ b/common/src/main/scala/uk/ac/wellcome/platform/archive/common/storage/services/SizeFinder.scala
@@ -1,29 +1,44 @@
 package uk.ac.wellcome.platform.archive.common.storage.services
 
 import com.amazonaws.services.s3.AmazonS3
-import uk.ac.wellcome.storage.ObjectLocation
+import com.amazonaws.services.s3.model.AmazonS3Exception
+import uk.ac.wellcome.storage.s3.S3Get
 import uk.ac.wellcome.storage.store.memory.MemoryStore
-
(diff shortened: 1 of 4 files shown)


common/src/main/scala/uk/ac/wellcome/platform/archive/common/storage/services/SizeFinder.scala
```diff
@@ -1,29 +1,51 @@
...
-  override def getSize(location: ObjectLocation): Try[Long] = Try {
-    memoryStore.entries
-      .getOrElse(location, throw new Throwable(s"No such entry $location!"))
-      .length
-  }
+  override def getSize(location: ObjectLocation): Either[ReadError, Long] =
+    memoryStore.entries.get(location) match {
+      case Some(entry) => Right(entry.length)
+      case None =>
+        Left(DoesNotExistError(new Throwable(s"No such entry $location!")))
```
kenoir, line 25: I feel like there should be some way of identifying these errors as from the size finder, rather than recycling them from the storage service. Maybe compose a sizefinder trait in with all the errors you want? Or just copy paste?

alexwlchan: You reckon? I feel it's conceptually similar to what you'd get from a Readable instance: this is a `Readable[ObjectLocation, Long]`, so the storage errors seem to fit.

PR #597, opened 2020-06-18
Use latest version of storage libs

kenoir:
For https://github.com/wellcomecollection/platform/issues/4397

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -33,14 +33,14 @@ def setupProject(
 // Temporarily commented out until https://github.com/wellcometrust/platform/issues/3806
 // In order to access our libraries in S3 we need to set the following:
 
-//s3CredentialsProvider := { _ =>
-//  val builder = new STSAssumeRoleSessionCredentialsProvider.Builder(
-//    "arn:aws:iam::760097843905:role/platform-read_only",
-//    UUID.randomUUID().toString
-//  )
-//
-//  builder.build()
-//}
+s3CredentialsProvider := { _ =>
+  val builder = new STSAssumeRoleSessionCredentialsProvider.Builder(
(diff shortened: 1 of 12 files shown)


build.sbt
```diff
@@ -33,14 +33,14 @@ def setupProject(
...
-//  builder.build()
-//}
+s3CredentialsProvider := { _ =>
+  val builder = new STSAssumeRoleSessionCredentialsProvider.Builder(
+    "arn:aws:iam::760097843905:role/platform-read_only",
+    UUID.randomUUID().toString
+  )
+
+  builder.build()
+}
```
alexwlchan, line 43: Ahem.

common/src/main/scala/uk/ac/wellcome/platform/archive/common/storage/services/SizeFinder.scala
```diff
@@ -28,13 +30,17 @@ class MemorySizeFinder(
...
-    val result =
-      S3Get
-        .get(location, maxRetries = 3) { location: ObjectLocation =>
-          s3Client
-            .getObjectMetadata(location.namespace, location.path)
-            .getContentLength
-        }
+
+    // TODO: Consider this being a "metadata readable"
+    // note this used to retry
```
alexwlchan, line 35: 😬

kenoir: What do you reckon? Ensure we maintain retries here?

alexwlchan: If we can, yeah, otherwise the register could get flaky.

PR #598, opened 2020-06-19
Use version 11.4.0 of scala-libs

kenoir:


common/src/main/scala/uk/ac/wellcome/platform/archive/common/config/builders/HTTPServerBuilder.scala
```diff
@@ -8,9 +8,9 @@ import uk.ac.wellcome.typesafe.config.builders.EnrichConfig._
 
 object HTTPServerBuilder {
   def buildHTTPServerConfig(config: Config): HTTPServerConfig = {
-    val host = config.required[String]("http.host")
-    val port = config.required[Int]("http.port")
```
alexwlchan, line 12: The fact this has never thrown makes me suspicious.

replica_aggregator/src/main/scala/uk/ac/wellcome/platform/storage/replica_aggregator/Main.scala
```diff
@@ -73,11 +73,11 @@ object Main extends WellcomeTypesafeApp {
       // TODO: Make this configurable
       replicaCounter = new ReplicaCounter(
         expectedReplicaCount =
-          config.required[String]("aggregator.expected_replica_count").toInt
+          config.requireString("aggregator.expected_replica_count").toInt
```
alexwlchan, line 76: While we're updating this config code, can we ditch the TODO comment above?

common/src/main/scala/uk/ac/wellcome/platform/archive/common/config/builders/HTTPServerBuilder.scala
kenoir: I don't like that I can't test this method either!

PR #599, opened 2020-06-22
Split the BagVerifier class into distinct steps

alexwlchan:
This opens the door to composing individual steps in other ways later.

Closes https://github.com/wellcomecollection/platform/issues/4593


bag_verifier/src/main/scala/uk/ac/wellcome/platform/archive/bagverifier/verify/steps/Step.scala
```diff
@@ -0,0 +1,18 @@
+package uk.ac.wellcome.platform.archive.bagverifier.verify.steps
+
+trait Step {
+  case class BagVerifierError(
```
kenoir, line 4: I might but this in a companion object so they are referenceable outside this trait without instantiating it.

bag_verifier/src/main/scala/uk/ac/wellcome/platform/archive/bagverifier/verify/steps/Step.scala
```diff
@@ -0,0 +1,18 @@
+package uk.ac.wellcome.platform.archive.bagverifier.verify.steps
+
+trait Step {
```
kenoir, line 3: 👍

bag_verifier/src/main/scala/uk/ac/wellcome/platform/archive/bagverifier/verify/steps/Step.scala
alexwlchan: Good suggestion, doing that.

(At which point the Step trait becomes so small it collapses in on itself, so it's out.)

PR #600, opened 2020-06-23
Modify SizeFinder to take an S3-specific location

alexwlchan:

Migrating the whole codebase in one go is way too big a job, and will either fail or leave us weeks of refactoring and clean-up afterwords.

What I’m instead going to do is migrate it gradually – create the new classes in the storage-service codebase, and work out all our uses of the old classes one-by-one. This makes each patch smaller and so I can keep the quality up, and I’m dealing with a much smaller number of compiler errors each time.

We'll only update the library when it's all migrated.


common/src/main/scala/uk/ac/wellcome/storage/store/NewRetryableReadable.scala
```diff
@@ -0,0 +1,36 @@
+package uk.ac.wellcome.storage.store
+
+import grizzled.slf4j.Logging
+import uk.ac.wellcome.storage.{Identified, ReadError, RetryOps}
+
+import scala.util.{Failure, Success, Try}
+
+trait NewRetryableReadable[Ident, T] extends Readable[Ident, T] with Logging {
```
alexwlchan, line 8: This is RetryableReadable with an extra type parameter.

PR #606, opened 2020-06-24
Use latest version of scala-libs > 14.0 - changes VHS models

kenoir:
This will require a migration of data held in the storage manifest VHS dynamo table.


kenoir:
> Gonna retract my review for now as I think there needs to be much more changes for this to work?

Quite possibly - this is me getting in a draft PR early (and getting travis to run the tests while i'm travelling).

warrd:
:+1: no probs

PR #607, opened 2020-07-01
Use the new location types inside the bag reader

alexwlchan:

### DIFF
--- a/bag_register/src/main/scala/uk/ac/wellcome/platform/archive/bag_register/services/Register.scala
+++ b/bag_register/src/main/scala/uk/ac/wellcome/platform/archive/bag_register/services/Register.scala
@@ -4,12 +4,21 @@ import java.time.Instant
 
 import grizzled.slf4j.Logging
 import uk.ac.wellcome.platform.archive.bag_register.models.RegistrationSummary
-import uk.ac.wellcome.platform.archive.bag_tracker.client.{BagTrackerClient, BagTrackerCreateError}
-import uk.ac.wellcome.platform.archive.common.bagit.models.{BagVersion, ExternalIdentifier}
+import uk.ac.wellcome.platform.archive.bag_tracker.client.{
+  BagTrackerClient,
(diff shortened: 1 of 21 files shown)


bag_register/src/main/scala/uk/ac/wellcome/platform/archive/bag_register/services/Register.scala
```diff
@@ -19,13 +19,16 @@ import uk.ac.wellcome.platform.archive.common.storage.services.{
   BadFetchLocationException,
   StorageManifestService
 }
+import uk.ac.wellcome.storage.{Location, ObjectLocationPrefix, Prefix}
 
 import scala.concurrent.{ExecutionContext, Future}
 
-class Register(
-  bagReader: BagReader,
+class Register[BagLocation <: Location, BagPrefix <: Prefix[BagLocation]](
```
kenoir, line 26: Must we have the type bounds here, or does it work with just the base type?

alexwlchan: I think so, because otherwise the compiler can't be sure the type bounds on BagReader are satisfied.

common/src/main/scala/uk/ac/wellcome/platform/archive/common/storage/services/memory/MemorySizeFinder.scala
```diff
@@ -7,9 +7,11 @@ import uk.ac.wellcome.storage.store.memory.MemoryStore
 class MemorySizeFinder[Ident](
   memoryStore: MemoryStore[Ident, Array[Byte]]
 ) extends SizeFinder[Ident] {
-  override def get(id: Ident): ReadEither =
+  override def get(id: Ident): ReadEither = {
+    println(s"@@AWLC getting id = $id")
```
kenoir, line 11: Needs removing I think.

PR #608, opened 2020-07-01
Split SourceLocation into per-provider models

alexwlchan:
Follows #607, part of wellcomecollection/platform#4596

When you start an ingest, you supply a SourceLocation. This could be S3 or Azure, and until now it's held a generic ObjectLocation. Changes:

* Ingests API: you can only upload a bag with the S3 provider. Anything else will be rejected immediately, because the unpacker doesn't support it.
* Strongly type the SourceLocation model with the appropriate provider/location model. We can’t try to unpack an S3 bag from an Azure location, or vice versa.
* Store a generic SourceLocation on the Ingest model. If we never add Azure support, it's just an extra `"type"` parameter in a few places; but we're set up to support it if we ever do.

This doesn't change the external interfaces – the ingests API or the indexed models. We will need to migrate the ingests table, so I’ll deploy this and look at that once this pull request is merged.

### DIFF
--- a/display/src/main/scala/uk/ac/wellcome/platform/archive/display/DisplayLocation.scala
+++ b/display/src/main/scala/uk/ac/wellcome/platform/archive/display/DisplayLocation.scala
@@ -3,7 +3,7 @@ package uk.ac.wellcome.platform.archive.display
 import io.circe.generic.extras.JsonKey
 import uk.ac.wellcome.platform.archive.common.ingests.models._
 import uk.ac.wellcome.platform.archive.common.storage.models.StorageLocation
-import uk.ac.wellcome.storage.{AzureBlobItemLocationPrefix, Prefix, S3ObjectLocationPrefix}
+import uk.ac.wellcome.storage.{AzureBlobItemLocationPrefix, S3ObjectLocationPrefix}
 
 case class DisplayLocation(
   provider: DisplayProvider,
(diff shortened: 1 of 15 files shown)


display/src/main/scala/uk/ac/wellcome/platform/archive/display/DisplayLocation.scala
```diff
@@ -12,19 +15,37 @@ case class DisplayLocation(
...
-    )
+    location match {
+      case S3SourceLocation(prefix) =>
+        DisplayLocation(
+          provider = DisplayProvider(location.provider),
+          bucket = prefix.bucket,
+          path = prefix.keyPrefix
+        )
+
+      case AzureBlobSourceLocation(prefix) =>
```
kenoir, line 42: I think we should check whether this will affect clients of the storage APIs (i.e. DDS).

alexwlchan: I don't think so – they’ll receive the same block of JSON as before, but whether you construct it from bucket/key or container/name depends on the provider. If I've done it right, this should be an entirely internal change that clients don't notice.

alexwlchan: It occurs to me we also don't need this change, now I added namespace/path methods to the Location model.

PR #610, opened 2020-07-01
Verify that the verifier checks the URI scheme in fetch.txt

alexwlchan:
I wasn't sure if it did, and stumbled across it while doing ObjectLocation work. Added a test that, yes, it really does the right thing.


alexwlchan:
Travis has passed but not shown a green tick.

PR #618, opened 2020-07-02
Use the new-style locations throughout the bag register

alexwlchan:
Final part of https://github.com/wellcomecollection/platform/issues/4596 for today.


bag_register/src/main/scala/uk/ac/wellcome/platform/archive/bag_register/services/StorageManifestException.scala
```diff
@@ -0,0 +1,7 @@
+package uk.ac.wellcome.platform.archive.bag_register.services
+
+class StorageManifestException(message: String)
```
kenoir, line 3: 👏

PR #623, opened 2020-07-06
Create the notion of the split S3BagVerifier

alexwlchan:
Heading towards https://github.com/wellcomecollection/platform/issues/4594

### DIFF
--- a/bag_verifier/src/main/scala/uk/ac/wellcome/platform/archive/bagverifier/services/BagVerifier.scala
+++ b/bag_verifier/src/main/scala/uk/ac/wellcome/platform/archive/bagverifier/services/BagVerifier.scala
@@ -201,8 +201,9 @@ trait BagVerifier[BagLocation <: Location, BagPrefix <: Prefix[BagLocation]]
     }
 }
 
-trait UnpackedBagVerifier[BagLocation <: Location, BagPrefix <: Prefix[BagLocation]]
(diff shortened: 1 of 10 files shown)


bag_verifier/src/main/scala/uk/ac/wellcome/platform/archive/bagverifier/services/BagVerifier.scala
```diff
@@ -200,3 +200,35 @@ trait BagVerifier[BagLocation <: Location, BagPrefix <: Prefix[BagLocation]]
...
+    root: BagPrefix,
+    space: StorageSpace,
+    externalIdentifier: ExternalIdentifier
+  ): Try[IngestStepResult[VerificationSummary]] =
+    verifyStandaloneBag(
+      ingestId = ingestId,
+      root = root,
+      space = space,
+      externalIdentifier = externalIdentifier
+    )
```
alexwlchan, line 235: The idea is that eventually this `verifyBag` method would take an extra argument `srcBagRoot`, and it would compare the `tagmanifest-sha256.txt` file before checking the checksums in the actual bag.

Because the function signatures are eventually going to diverge, that means divergence in `Main` and in `BagVerifierWorker`, and it's worth discussing how we do that before going on further.

alexwlchan:
Closing in favour of #630.

PR #630, opened 2020-07-08
[WIP] Verify tag manifests match in bag verifier

alicefuzier:
For https://github.com/wellcomecollection/platform/issues/4594


bag_verifier/src/test/scala/uk/ac/wellcome/platform/archive/bagverifier/services/BagVerifierTestCases.scala
```diff
@@ -304,6 +306,44 @@ trait BagVerifierTestCases[BagLocation <: Location, BagPrefix <: Prefix[
...
+
+        val result = ingestStep.success.get
+
+        result shouldBe a[IngestFailed[_]]
+        result.summary shouldBe a[VerificationIncompleteSummary]
+
+        result.maybeUserFacingMessage shouldNot be(defined)
+      }
+    }
+  }
```
alexwlchan, line 345: Really nice, you've got the hang of these tests. :)

bag_verifier/src/test/scala/uk/ac/wellcome/platform/archive/bagverifier/services/BagVerifierTestCases.scala
```diff
@@ -663,3 +686,58 @@ trait BagVerifierTestCases[BagLocation <: Location, BagPrefix <: Prefix[
...
+      BagPrefix,
+      Namespace
+    ] {
+  override def createBagRoot(
+    bagRoot: BagPrefix,
+    scrBagRoot: Option[BagPrefix]
+  ): ReplicatedBagVerifyContext[BagLocation, BagPrefix] =
+    ReplicatedBagVerifyContext(bagRoot, scrBagRoot.getOrElse(bagRoot))
+
+  it("fails a bag if it doesn't match original tag manifest") {
```
alexwlchan, line 707: Other test cases that would be useful:

* If the original and replicated `tagmanifest-sha256.txt` match, the bag is verified successfully. (Previously this was tested implicitly in the replicator – we'd know it worked if the other replicator tests passed.)

* What if we can't find the source bag at all?

bag_verifier/src/main/scala/uk/ac/wellcome/platform/archive/bagverifier/builder/BagVerifierWorkerBuilder.scala
```diff
@@ -0,0 +1,115 @@
...
+    ingestUpdater: IngestUpdater[SNSConfig],
+    outgoingPublisher: OutgoingPublisher[SNSConfig]
+  )(
+    implicit s3: AmazonS3,
+    mc: MetricsMonitoringClient,
+    as: ActorSystem,
+    sc: SqsAsyncClient
+  ) = {
+    val isReplicatedBagMode =
+      config.getBoolean("bag-verifier.mode.replicated_bag")
```
alexwlchan, line 40: Right now we run two verifiers:

* Verifying the unpacked, standalone bag in S3
* Verifying the replicated bag in S3

Soon we’ll add:

* Verifying the replicated bag in Azure

I guess at that point, we could add a third value to this flag (and make it getString()) to determine the verifier behaviour? For now, maybe tweak the name so it’s a boolean? `bag-verifier.mode.is-replica`.

Something like `unpacked-s3`, `replica-s3`, `replica-azure`?

I like the flexibility it gives us. 👍

bag_verifier/src/test/scala/uk/ac/wellcome/platform/archive/bagverifier/services/BagVerifierTestCases.scala
alicefuzier: yep, I'll add them

bag_verifier/src/main/scala/uk/ac/wellcome/platform/archive/bagverifier/services/BagVerifier.scala
```diff
@@ -16,95 +16,100 @@ import uk.ac.wellcome.storage.{Location, Prefix}
 
 import scala.util.Try
 
-trait BagVerifier[BagLocation <: Location, BagPrefix <: Prefix[BagLocation]]
-    extends Logging
+sealed trait BagVerifyContext[BagLocation <: Location, BagPrefix <: Prefix[
+  BagLocation
+]] {
+  val root: BagPrefix
+}
```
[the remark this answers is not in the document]
alicefuzier: Uhm that doesn't seem to work. When I do this and the try to instantiate one, it says that it doesn't match the type boundaries. I think that's because `Prefix` is not covariant. I think I'll leave it as is unless you have strong opinions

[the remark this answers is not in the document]
alexwlchan: Sure; it'd be a nice cleanup if it works but not essential.

bag_verifier/src/test/scala/uk/ac/wellcome/platform/archive/bagverifier/services/BagVerifierTestCases.scala
alicefuzier: So, the first test is kinda done in that the `ReplicatedBagTestCases` extends `BagVerifierTestCases` and there first test is the passing one where they match. I'm saying _kinda_ done because the way it's done is the path of the bag and the path of the source are the same so of course the tagmanifest match

alexwlchan:
I got an email “Alice pushed 0 commits” which seems to have upset the Travis web hooks.

PR #635, opened 2020-07-14
Run a replicator that can write to Azure

alexwlchan:
A first step towards https://github.com/wellcomecollection/platform/issues/4587

We need to do something to turn this off in prod until we have the rest of the pieces in place, but it works! We wrote some blobs to staging!

### DIFF
--- a/bag_replicator/src/main/scala/uk/ac/wellcome/platform/archive/bagreplicator/Main.scala
+++ b/bag_replicator/src/main/scala/uk/ac/wellcome/platform/archive/bagreplicator/Main.scala
@@ -26,6 +26,11 @@ import uk.ac.wellcome.platform.archive.common.config.builders.{
   OperationNameBuilder,
   OutgoingPublisherBuilder
 }
+import uk.ac.wellcome.platform.archive.common.ingests.models.{
+  AmazonS3StorageProvider,
+  AzureBlobStorageProvider,
+  StorageProvider
+}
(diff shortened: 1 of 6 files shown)


bag_replicator/src/main/scala/uk/ac/wellcome/platform/archive/bagreplicator/Main.scala
```diff
@@ -70,7 +72,16 @@ object Main extends WellcomeTypesafeApp {
     val lockingService =
       new DynamoLockingService[IngestStepResult[BagReplicationSummary[_]], Try]()
 
-    val replicator = new S3Replicator()
+    val replicator = config.requireString("bag-replicator.provider") match {
+      case "amazon-s3"          => new S3Replicator()
+      case "azure-blob-storage" =>
```
kenoir, line 77: You could use: https://github.com/wellcomecollection/scala-libs/blob/master/storage_typesafe/src/main/scala/uk/ac/wellcome/storage/typesafe/AzureBlobServiceClientBuilder.scala#L7

alexwlchan: Ah, I’ve not committed the change, but connectionString turns out to be the wrong thing for SAS tokens – this should be `endpoint()`.

But good shout, we should definitely get this into the libs.

PR #636, opened 2020-07-15
Define our Azure storage containers in Terraform

alexwlchan:
Closes https://github.com/wellcomecollection/platform/issues/4661

This patch creates a pair of Azure storage accounts (`wecostorageprod` and `wecostoragestaging`), which contain two containers for our replicas, named to match the existing set:

* `wellcomecollection-storage`
* `wellcomecollection-storage-replica-ireland`
* **`wellcomecollection-storage-replica-netherlands`**

<br>

* `wellcomecollection-storage-staging`
* `wellcomecollection-storage-staging-replica-ireland`
* **`wellcomecollection-storage-staging-replica-netherlands`**

Plus a rule that moves blobs to the Archive tier after 30 days.

Notes:

* For @alicefuzier: in the storage service, Terraform resources are split into "critical" and "stack". All the irreplaceable databases live in the critical stack, which we change infrequently – so we don't have to worry about inadvertently deleting a key bucket when deploying a new ECS service or something. (I can't remember if we have this pattern in the catalogue.)

* For @jtweed: I've selected **Locally Redundant Storage** replication and the **West Europe (Netherlands)** region based on past conversations. Are those both correct?


jtweed:
Yes I believe so. That’s still redundant within the facility, which is fine for our copy of last resort. Europe North is also half the price of UK South.

PR #648, opened 2020-07-22
Ditch the BagReplicationRequest class

alexwlchan:
Another bit of refactoring as part of https://github.com/wellcomecollection/platform/issues/4596

Previously we had ReplicationRequest and BagReplicationRequest. The former was (srcPrefix, dstPrefix); the latter also tracked whether the replica was primary/secondary.

Arguably this shouldn't be passed around inside BagReplicatorWorker at all – it's only interesting when we go to write the result payload, and send it on to the replica aggregator (which needs to distinguish between primary/secondary replicas). So this patch introduces a new, simpler `ReplicaType` model, which is part of the replicator config. We use that to construct the payload, but we don't pass it around elsewhere.

This doesn't change the replicator output, but it gets us in a position where we can start sending the new-style locations to the replica aggregator in the next patch.


bag_replicator/src/main/scala/uk/ac/wellcome/platform/archive/bagreplicator/models/ReplicaType.scala
```diff
@@ -0,0 +1,15 @@
...
+//  - The primary replica is the "warm" copy, intended for frequent access
+//    e.g. an S3 bucket with Standard-IA
+//
+//  - Secondary replicas are the "cold" copies, that give us an extra layer
+//    of backup but aren't intended for frequent access
+//    e.g. an S3 Glacier bucket
+
+sealed trait ReplicaType
+
+case object PrimaryReplica extends ReplicaType
```
kenoir, line 14: 👏

PR #649, opened 2020-07-22
Pass a new-style ReplicaLocation between the replicator and the aggregator

alexwlchan:
Follows #648, part of https://github.com/wellcomecollection/platform/issues/4596

Both apps are using ReplicaLocation internally, so we can strip off a translation layer and pass it between them.


bag_replicator/src/main/scala/uk/ac/wellcome/platform/archive/bagreplicator/services/BagReplicatorWorker.scala
```diff
@@ -89,8 +89,7 @@ class BagReplicatorWorker[
             .toReplicaLocation(
               provider = destinationConfig.provider,
               replicaType = destinationConfig.replicaType
-            )
-            .toStorageLocation,
+            ),
```
kenoir, line 92: 🕺

PR #650, opened 2020-07-22
Remove an unused method from the bag root finder tests

alexwlchan:
Snip, snip, snip.


alexwlchan:
Passing in Travis, but the commit status hasn't updated. Merging.

PR #654, opened 2020-07-23
Use new .wellcome_project format

kenoir:
Use new project format & use new version of deploy tool in build.


alexwlchan:
Merging so I can deploy from master.

PR #655, opened 2020-07-23
The bag register only accepts a primary S3 location

alexwlchan:
Another part of https://github.com/wellcomecollection/platform/issues/4596; spun out of https://github.com/wellcomecollection/storage-service/pull/653

The reality is that the bag register only deals with bags whose primary copy is in S3, and the pipeline is woefully under-equipped to handle anything else. It makes our life easier if we just accept this, and make it an explicit part of the codebase – if we ever want to register bags in other locations, then we can come back and make it nice and generic, but right now it's easier to be S3 specific.

Plus some other test refactorings.


kenoir:
Why is this easier? Definitely less code here and simpler - but does this enable more change later or is this only because we don't register things in Azure _right now_. If the latter, it's arguable that this change is not necessary.

alexwlchan:
It's mostly a cleanup, and recognising that the S3StorageManifestService (the thing we care about) wasn't actually being tested.

It makes things easier because we can assert that yes, the primary location of the bag is an S3ObjectLocation, rather than treating it as a generic BagLocation and writing a bunch of methods around it. In particular, the twiddling around the versioned paths gets awkward when you do it in a generic way. I tried a bunch of approaches; taking out the type parameters is the simplest approach.

It'd be nice to come back and make it generic *at some point*, but for the sake of expediency and finishing off the ObjectLocation disambiguation, it was simpler to become less generic for a bit.

PR #657, opened 2020-07-24
Add a script for migrating StorageLocations

alexwlchan:
For https://github.com/wellcomecollection/platform/issues/4596

I ran this script over the storage manifest VHS; it updated the StorageLocation in the manifests and wrote that to a new VHS. Then I updated the apps to point at the new VHS instance.

This change has been run and applied in prod.


alexwlchan:
Merging this to get the new Terraform definitions in so I can apply #659.

PR #658, opened 2020-07-24
Use the new version of scala-libs

alexwlchan:


bag_replicator/src/main/scala/uk/ac/wellcome/platform/archive/bagreplicator/replicator/Replicator.scala
```diff
@@ -99,8 +100,11 @@ trait Replicator[DstLocation <: Location, DstPrefix <: Prefix[DstLocation]]
       case Right(_) =>
         ReplicationSucceeded(summary.complete)
 
-      case Left(err: TransferResult) =>
-        ReplicationFailed(summary.complete, err.e)
+      case Left(err: PrefixTransferFailure) =>
+        ReplicationFailed(
+          summary.complete,
+          e = new Throwable(s"Prefix transfer failed: $err")
```
alicefuzier, line 106: minor but why this? don't we lose the stack trace this way? Might not be a problem but you could wrap the original exception in the new one and retain that. Not sure if you can do that with Throwables though

PR #662, opened 2020-07-28
12 task roles for iiif builder configs

tomcrane:
This works, and it's an extension of what we have before.
But 12 identifiers here feels a bit odd.

Is this acceptable now, or should we look to find a way to put all these different services/configurations behind one role that presents itself to the staging and production storage buckets?


tomcrane:
Partner of https://github.com/wellcomecollection/iiif-builder-infrastructure/pull/15

alexwlchan:
I suspect this is superseded by https://github.com/wellcomecollection/storage-service/pull/664, so closing, but happy to re-review if that turns out not to work.

donaldgray:
#664 has been tested and is working as expected.

PR #663, opened 2020-07-29
Give S3 access to the whole Digirati account, rather than specific roles

alexwlchan:


alexwlchan:
Closing in favour of #664

PR #670, opened 2020-08-03
Add a script to find all the bags without Azure replicas

alexwlchan:
A first step towards https://github.com/wellcomecollection/platform/issues/4589

It generates instances of VersionedBagRootPayload for any bag in the replicas table that doesn't have an Azure replica:

```json
{
  "context": {
    "ingestId": "636ae000-9ce7-4f39-a350-cf2ce9f31428",
    "ingestType": {
      "id": "update"
    },
    "storageSpace": "born-digital-accessions",
    "externalIdentifier": "WT/B/9/2/2",
    "ingestDate": "2019-12-13T10:46:41Z"
  },
  "bagRoot": {
    "bucket": "wellcomecollection-storage-staging",
    "keyPrefix": "born-digital-accessions/WT/B/9/2/2/v7"
  },
  "version": 7,
  "type": "VersionedBagRootPayload"
}
```

When we verify, we'll want to look at the actual storage manifests to ensure the new replicas were registered, but walking the manifests table is way more expensive. The theory is that we could run this repeatedly until we see a complete set of replicas.


scripts/migrations/2020-08-create_azure_replicas.py
```diff
@@ -0,0 +1,161 @@
...
+        size = s3_client.head_object(Bucket=bucket, Key=key)["ContentLength"]
+        s3_fragment = (
+            s3_client.get_object(Bucket=bucket, Key=key, Range=f"bytes={size - 51}-")[
+                "Body"
+            ]
+            .read()
+            .decode("utf8")
+        )
+
+        return re.match(r'^"ingestId": "([0-9a-f-]{36})"}$', s3_fragment).group(1)
```
kenoir, line 87: fancy!

alexwlchan: They say necessity is the mother of invention. I say a slow internet connection is the mother of bad code hacks. 😜

PR #676, opened 2020-08-07
Add an explanation of bag data movements + terminology

alexwlchan:


replifier/README.md
```diff
@@ -0,0 +1,97 @@
...
+
+## How data moves through the storage service
+
+![](data_movements.png)
+
+1.  The user uploads a bag to the **user uploads bucket**.
+
+2.  The bag unpacker unpacks the user-uploaded bag into the **unpacked bags bucket**.
+
+    We want it inside our AWS account as soon as possible (so it can't be deleted under our feet).
```
alicefuzier, line 18: stupid question: does that mean that the `user uploads bucket` is outside our AWS account?

replifier/README.md
```diff
@@ -0,0 +1,97 @@
...
+    This re-checks that the bag is valid, and confirms that it's the same as the originally unpacked bag.
+
+In theory, each of these buckets could be backed by a different storage provider.
+In practice:
+
+*   We only support user uploads from S3.
+*   We unpack bags into a bucket in S3.
+*   The replicas can be a mix of providers: currently we have one in S3, one in S3 Glacier, one in Azure.
+
+At time of writing (July 2020), our codebase has hard-coded assumptions that the user uploads and unpacked bags buckets are both in S3.
```
alicefuzier, line 47: Hard to believe, but we are already in August 🙂

replifier/README.md
alexwlchan: Not stupid – right now it’s inside our AWS estate but not the same account (we have upload buckets in the workflow/digitisation accounts), but in theory it could be anywhere. I'll clarify.

alicefuzier: Ignore it, I was reading the non rendered version so wasn't seeing the picture with the clear boundary of our account

replifier/README.md
alexwlchan: 😱

PR #669, opened 2020-08-03
Add AzureBagVerifier

alicefuzier:

PR #680, opened 2020-08-11
Add an implementation of AzureBagReader

alexwlchan:
Part of https://github.com/wellcomecollection/platform/issues/4595. Extracted as a standalone piece from https://github.com/wellcomecollection/storage-service/pull/669

This was a single, standalone commit that can be merged on its own.


common/src/test/scala/uk/ac/wellcome/platform/archive/common/bagit/services/azure/AzureBagReaderTest.scala
```diff
@@ -0,0 +1,34 @@
...
+import uk.ac.wellcome.storage.store.azure.AzureTypedStore
+
+class AzureBagReaderTest extends BagReaderTestCases[
+  Unit,
+  Container,
+  AzureBlobLocation,
+  AzureBlobLocationPrefix
+]
+  with AzureBagBuilder {
+  override def asLocation(root: AzureBlobLocationPrefix, path: String): AzureBlobLocation = root.asLocation(path)
```
alexwlchan: Note: this method can be pushed down into BagReaderTestCases, because this is now a required method on Prefix[Location], but let's do that separately.

PR #691, opened 2020-08-12
Allow other types of storage in BagVerifierWorker

alicefuzier:


bag_verifier/src/test/scala/uk/ac/wellcome/platform/archive/bagverifier/fixtures/BagVerifierFixtures.scala
```diff
@@ -66,18 +69,22 @@ trait BagVerifierFixtures
       }
     }
 
-  def withReplicaBagVerifierWorker[R](
+  def withReplicaBagVerifierWorker[BagLocation <: Location, BagPrefix <: Prefix[
```
alexwlchan, line 72: Possibly daft question, but what are these type parameters for? Are they adding anything?

alicefuzier: Ah yes. I've created two of these in another branch. One for Azure and one for S3.I'm going to merge as is as it's not staying like that anyway

PR #693, opened 2020-08-14
Don't fail a verification if we can't write tags to Azure

alexwlchan:
Because we know there are issues with the fixity tag we write to S3 if we write it to Azure: https://github.com/wellcomecollection/platform/issues/4730

We're not ignoring all errors in tagging (in particular, if we have a tagging issue in S3, it should still bubble up as a failure), nor are we injecting config deep into the heart of the verifier to decide whether to tag. It always tries to tag, and sometimes ignores the failure.

(So if Azure ever change their mind about what metadata is allowed, this code will quietly start working.)

Still needs a test that it actually lets an Azure tagging failure pass through.


alexwlchan:
Closing as a rejected approach.

PR #701, opened 2020-08-18
Azure Backfill infra

alicefuzier:


terraform/modules/stack/main.tf
```diff
@@ -569,7 +569,7 @@ module "replica_aggregator" {
     queue_url              = module.replica_aggregator_input_queue.url
     outgoing_topic_arn     = module.replica_aggregator_output_topic.arn
     ingest_topic_arn       = module.ingests_topic.arn
-    metrics_namespace      = local.bag_register_service_name
+    metrics_namespace      = local.replica_aggregator_service_name
```
alexwlchan, line 572: 😅 nice catch

terraform/modules/stack/iam_role_policy.tf
```diff
@@ -181,3 +181,29 @@ resource "aws_iam_role_policy" "notifier_metrics" {
...
+  role   = module.bag_register_backfill.task_role_name
+  policy = data.aws_iam_policy_document.replica_primary_readonly.json
+}
+
+resource "aws_iam_role_policy" "bag_register_backfill_metrics" {
+  role   = module.bag_register_backfill.task_role_name
+  policy = data.aws_iam_policy_document.cloudwatch_putmetrics.json
+}
+
+resource "aws_iam_role_policy" "azure_replicator_replica_primary_readonly" {
```
alexwlchan, line 206: Possibly worth a TODO: delete on this one? The backfill roles we'll remember to clean up because the services will get deleted; this one won't.

alicefuzier: The idea was to add all of this backfill stuff at the end of each file so that we can just go ahead and delete everything under a `# backfill` comment. I'll add a todo in each one so we know and it's not just in my head

PR #708, opened 2020-08-21
Fix verifier listing "directories"

alicefuzier:


bag_verifier/src/main/scala/uk/ac/wellcome/platform/archive/bagverifier/services/BagVerifier.scala
```diff
@@ -238,4 +241,6 @@ trait BagVerifier[BagContext <: BagVerifyContext[BagPrefix], BagLocation <: Loca
           maybeUserFacingMessage = Some(userFacingMessage)
         )
     }
+
+  def addTrailingSlash(prefix: BagPrefix): BagPrefix
```
alexwlchan, line 245: One more minor nit: this doesn't always add a slash; sometimes it leaves it unmodified.

Maybe `ensureTrailingSlash`?

alexwlchan (changes requested):
Having read this more carefully: what if the verifier gets asked to verify a prefix that already ends with a trailing slash?

I think that `addTrailingSlash` method should become `ensureTrailingSlash`, and maybe we push that functionality into a shared object – we'll want to ensure trailing slashes in other places.

bag_verifier/src/main/scala/uk/ac/wellcome/platform/archive/bagverifier/services/BagVerifier.scala
alicefuzier: not really, this gets called only after the check of whether the trailing slash already contains the slash

alexwlchan: Duh. Oh yeah, ignore me.

alexwlchan (approved):
Ignore my last comment; I missed the check for the trailing slash.

bag_verifier/src/test/scala/uk/ac/wellcome/platform/archive/bagverifier/services/BagVerifierTestCases.scala
```diff
@@ -76,21 +76,21 @@ trait BagVerifierTestCases[Verifier <: BagVerifier[
   def withBag[R](
     space: StorageSpace,
     externalIdentifier: ExternalIdentifier,
-    bagBuilder: BagBuilder[BagLocation, BagPrefix, Namespace] = bagBuilder
-  )(testWith: TestWith[(Bucket, BagPrefix), R]): R =
+    bagBuilder: BagBuilder[BagLocation, BagPrefix, Namespace] = bagBuilder,
+    version: BagVersion = BagVersion(randomInt(from = 2, to = 10))
+  )(namespace: Namespace)(testWith: TestWith[(Bucket, BagPrefix), R]): R =
```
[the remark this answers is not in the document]
alicefuzier: yeah, although it took me a while to understand why the test that I was expecting to fail was in fact succeeding and that was because the namespace was hidden inside the `withBag` method. Maybe it's not all that bad that we make the namespace more explicit

bag_verifier/src/test/scala/uk/ac/wellcome/platform/archive/bagverifier/services/BagVerifierTestCases.scala
```diff
@@ -134,34 +134,79 @@ trait BagVerifierTestCases[Verifier <: BagVerifier[
...
+          verifySuccessCount(
+            fixityListResult.locations,
+            expectedCount = expectedFileCount
+          )
+      }
+    }
+  }
+
+  it(
+    "doesn't match locations in a namespace with same prefix but different directory"
```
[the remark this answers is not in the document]
alicefuzier: Ah ok, I tend to order them happy path first and error modes after but yeah I'll move it down

PR #709, opened 2020-08-22
Add an Azure replicator that uses the Put Block from URL API call

alexwlchan:
This is an experimental approach to https://github.com/wellcomecollection/platform/issues/4757.

It can be broken down and needs to be commented/documented properly before we can merge it, and some of the tests are failing.

The tl;dr: you can't upload a large InputStream as a new blob to Azure in one go. You have to upload it as a series of "blocks" using the "Put Block" APIs, then call "Put Block List", which stitches a series of blocks into a single blob. We have to write our own logic for splitting an S3 object into individual blocks, and ensuring they get concatenated successfully.

It would be better to use the "Put Block URL" from API, which drastically simplifies memory and bandwidth (and is probably much faster!), but that doesn't work with the Azurite container we use. I don't want all that logic for splitting/concatenating blobcks to go untested, so I created two implementations:

* AzureTransfer is a base trait that contains all the logic for splitting blobs (with a bit of it pushed out to BlobRangeUtil)
* It has two implementations: PutBlockTransfer (which we can test with Azurite) and PutBlockFromURLTransfer (which we can't, but is only a tiny implementation)

This minimises the amount of untested code while letting us upload much bigger objects.


bag_replicator/src/main/scala/uk/ac/wellcome/platform/archive/bagreplicator/storage/azure/BlobRangeUtil.scala
```diff
@@ -0,0 +1,39 @@
+package uk.ac.wellcome.platform.archive.bagreplicator.storage.azure
+
+import java.util.Base64
+
+import com.azure.storage.blob.models.BlobRange
+
+object BlobRangeUtil {
+  def getRanges(length: Long, blockSize: Long): Seq[BlobRange] =
+    ((0L to length - 1) by blockSize)
```
alicefuzier, line 9: i didn't know you could do this! Nice

alexwlchan: Yeah, new to me too! Much nicer than the icky Enumeration code I wrote elsewhere.

alicefuzier (approved):
Nice work! Really neat to extract the untestable bit into the `AzureTransfer` trait

kenoir:
Holy cow, this is quite an undertaking. I am surprised there is not an effective way provided by the Azure SDK to transfer large objects to Azure.

PR #714, opened 2020-08-25
Introduce a RangedReader and LargeStreamReader trait

alexwlchan:
Refactoring for https://github.com/wellcomecollection/platform/issues/4757 and https://github.com/wellcomecollection/platform/issues/4743

1. For unpacking large bags in S3, we have a "large stream reader" which:

    * Finds out the size of the object
    * Reads it in chunks of a fixed size
    * Stitches those chunks back together into an InputStream

    None of that logic is S3 specific, and we've gradually split most of it into distinct classes/services.

2. We're having issues verifying large objects in Azure – the InputStream times out midway through, and the whole verification fails.

    This hasn't been an issue in S3. Guesses as to why: cross-cloud latency makes things flaky, S3 is inherently more reliable.

What if we combined these ideas? 🤯 

My plan is to create an Azure implementation of the "large stream reader", and use that in the verifier. Hopefully that will improve the reliability somewhat, and allow us to verify the biggest of objects.

This PR just rearranges the existing code and classes so that an Azure implementation should be a simple drop-in, and ties together some refactoring and scala-libs improvements previously unused. No change in functionality.


alicefuzier (approved):
Nice fiddly work. And thanks for the comments. I don't think I would have been able to go through to the code without them

PR #715, opened 2020-08-25
Add an implementation of a LargeStreamReader for Azure

alexwlchan:
Follows https://github.com/wellcomecollection/storage-service/pull/714; for wellcomecollection/platform#4757 and wellcomecollection/platform#4743

PR #717, opened 2020-08-25
Add an implementation of a LargeStreamReader for Azure

alexwlchan:
https://github.com/wellcomecollection/storage-service/pull/715, but into the correct branch


alicefuzier (approved):
LOL, it took me a while to understand why I saw this PR twice

PR #718, opened 2020-08-25
Add script to update createdDate in backfill VHS

alicefuzier:


scripts/migrations/2020-08-update_migration_created_date.py
```diff
@@ -0,0 +1,65 @@
...
+    except KeyError:
+        try:
+            bucket = item["payload"]["bucket"]
+            key = item["payload"]["key"]
+        except:
+            raise RuntimeError(f"Find s3 bucket, key in {item}")
+        else:
+            return bucket, key
+    else:
+        return bucket, key
```
alexwlchan, line 29: Two notes:

* It feels like this would be a good opportunity to tidy up the VHS structure – and use `bucket`/`key` everywhere.
* In general, bare `except:` is a bad idea in Python – it captures _everything_, including a keyboard interrupt or out-of-memory error. Normally not what you want; it's better to be more specific with the errors you catch.

scripts/migrations/2020-08-update_migration_created_date.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+def put_vhs_json(bucket, key, content):
+    try:
+        s3object = s3.Object(bucket, key)
+        # s3object.put(
+        #     Body=(bytes(json.dumps(content).encode('UTF-8')))
+        # )
+    except Error as e:
+        raise RuntimeError(f"Error updating backfill vhs object {bucket} {key}: {e}")
```
alexwlchan: If I've read correctly, this will overwrite the existing S3 object?

When I've done migrations in the past, I written to a brand new object – that way, the before/after objects are both preserved. It's not useful long-term, but it avoids blatting information, e.g. if I muck up the script and accidentally write empty data to S3.

scripts/migrations/2020-08-update_migration_created_date.py
```diff
@@ -0,0 +1,67 @@
...
+            backfilled_item = dynamodb.Table(backfill_vhs_table).get_item(
+                Key={"id": id, "version": version}
+            )["Item"]
+        except KeyError:
+            print(f"Cannot find backfill storage manifest for {id}:{version}!!!!")
+        else:
+            backfilled_bucket = backfilled_item["payload"]["bucket"]
+            backfilled_key = backfilled_item["payload"]["key"]
+            backfilled_json = get_vhs_json(backfilled_bucket, backfilled_key)
+            backfilled_json["createdDate"] = created_date
```
alexwlchan, line 64: I think it's worth doing some other checks here of the storage manifest from the backfilled VHS.

e.g.

* They both have the same top-level JSON keys
* They both have the same values (createdDate and replicas aside)
* The new manifest has one `"location"` and two `"replicas"` – two in S3, one in Azure

scripts/migrations/2020-08-update_migration_created_date.py
```diff
@@ -0,0 +1,65 @@
...
+
+for item in scan_table(TableName=vhs_table):
+    id = item["id"]
+    version = item["version"]
+    bucket, key = get_bucket_key(item)
+    try:
+        vhs_content = get_vhs_json(bucket, key)
+    except:
+        raise RuntimeError(f"Cannot read s3 entry for {id}: {item}")
+    else:
```
alexwlchan, line 52: It might be worth a bit here that checks for three replicas in the manifest already in VHS, and if they're there, it skips copying the createdDate across. That way, you can step through it, one bag at a time if necessary.

That’s typically how I write this sort of script – I write it so that it skips redoing work that’s already been done, then I use `break` to do a single step at a time until I'm happy it works.

alexwlchan (approved):
Looks pretty good for a first pass. See suggestions inline, specifically:

* Rather than blatting existing S3 objects, I prefer to create a whole new object and update the pointer in VHS
* While we've got the storage manifest open, let's be picky and check it looks like we expect

scripts/migrations/2020-08-update_migration_created_date.py
alicefuzier: makes sense, thanks

scripts/migrations/2020-08-update_migration_created_date.py
alicefuzier: ok, will do. But if we update the pointer isn't it kind of the same thing? The old object will be lost anyway without anything referring to it won't it?

scripts/migrations/2020-08-update_migration_created_date.py
alicefuzier: the new VHS uses bucket/key everywhere already. It's the old one that is a mixture.

I'll avoid catching everything, thanks

scripts/migrations/2020-08-update_migration_created_date.py
alicefuzier: Actually, I'm not sure this is true. We have sent bags with Azure replication enabled but without the backfill infra in place. So there might be bags for which the current VHS has 3 replicas but the createdDate in the backfill one is still incorrect. I think it would be better to leave it doing the work more than once and not optimize in this instance. if the db is too big in prod we can always parallelize right?

scripts/migrations/2020-08-update_migration_created_date.py
```diff
@@ -0,0 +1,83 @@
...
+                items_added = diff.pop("iterable_item_added", None)
+                if values_changed and values_changed.keys()-["root['createdDate']"]:
+                    raise RuntimeError(f"Unexpected values changed in {values_changed} for {id}:{version}!")
+                if items_added and items_added.keys() - ["root['replicaLocations'][1]"]:
+                    raise RuntimeError(f"Unexpected values changed in {items_added} for {id}:{version}!")
+                if diff:
+                    raise RuntimeError(f"Unexpected values changed in {diff} for {id}:{version}!")
+
+            backfilled_json["createdDate"] = created_date
+            put_vhs_json(backfilled_bucket, backfilled_key, backfilled_json)
```
alexwlchan, line 83: This is still overwriting the existing object, right?

alexwlchan: Also… hmm, is the backfill bucket the appropriate place? It's got backfill in the name, which feels a bit icky. Maybe we should be pushing back into the live VHS (or its bucket)? 😕

alicefuzier: yes, I still have to change that...the syntax to update something in dynamo isn't incredibly friendly to me so trying to navigate that right now. 

Yeh, maybe if we are updating the VHS record anyway, it could be better updating the old one instead of the new one

scripts/migrations/2020-08-update_migration_created_date.py
```diff
@@ -0,0 +1,115 @@
...
+def get_vhs_json(id, version, bucket, key):
+    try:
+        return json.loads(s3.Object(bucket, key).get()["Body"].read().decode("utf-8"))
+    except ClientError as e:
+        record_error(id, version, f"Cannot read s3 entry for {id}: {item}: {e}")
+
+
+def put_vhs_json(id, version, bucket, content):
+    bb = uuid.uuid4()
+    key = f"{id}/{version}/{bb}"
```
alexwlchan, line 49:
```suggestion
    key = f"{id}/{version}/{bb}.json"
```

We now stick `.json` on the end of these keys, so it's slightly easier to deal with if you're browsing the VHS by hand.

alicefuzier: ah thanks, missed that

scripts/migrations/2020-08-update_migration_created_date.py
```diff
@@ -0,0 +1,118 @@
...
+        return json.loads(s3.Object(bucket, key).get()["Body"].read().decode("utf-8"))
+    except ClientError as e:
+        record_error(id, version, f"Cannot read s3 entry for {id}: {item}: {e}")
+
+
+def put_vhs_json(id, version, bucket, item, content):
+    filename = f"{uuid.uuid4()}.json"
+    key = f"{id}/{version}/{filename}"
+    item["payload"]["bucket"] = bucket
+    item["payload"]["key"] = key
```
alexwlchan, line 53: Possibly better:

```
item["payload"] = {"bucket": bucket, "key": key}
```

so a straggling `namespace` field will get purged.

scripts/migrations/2020-08-update_migration_created_date.py
```diff
@@ -0,0 +1,118 @@
...
+            record_error(id, version, f"Unexpected values changed in {diff}!")
+            return False
+    return True
+
+
+for item in scan_table(TableName=vhs_table):
+    id = item["id"]
+    version = item["version"]
+    bucket, key = get_bucket_key(item)
+    if bucket and key:
```
alexwlchan, line 96: So FWIW, I suspect this would explode if `get_bucket_key()` ever failed to find anything. It would return `None`, and this line would become

```python
bucket, key = None
```

which is a TypeError – you can’t unpack `None` into two variables.

I doubt it'll be an issue here, but a note for the next Python you write.

scripts/migrations/2020-08-update_migration_created_date.py
```diff
@@ -0,0 +1,118 @@
...
+        if vhs_content:
+            created_date = vhs_content["createdDate"]
+            backfilled_item = get_backfill_item(id, version)
+            if backfilled_item:
+                backfilled_bucket = backfilled_item["payload"]["bucket"]
+                backfilled_key = backfilled_item["payload"]["key"]
+                backfilled_json = get_vhs_json(
+                    id, version, backfilled_bucket, backfilled_key
+                )
+                if backfilled_json:
```
alexwlchan, line 107: Might be worth a line here:

```python
if backfilled_json == vhs_content:
    continue
```

so you won't keep writing the same manifests over and over.

alexwlchan (approved):

PR #729, opened 2020-09-09
Add an implementation of Azure tags in Dynamo

alexwlchan:
What if we stuck tags about Azure objects in a DynamoDB sidecar?

### DIFF
--- a/bag_verifier/src/main/scala/uk/ac/wellcome/platform/archive/bagverifier/fixity/azure/AzureDynamoTags.scala
+++ b/bag_verifier/src/main/scala/uk/ac/wellcome/platform/archive/bagverifier/fixity/azure/AzureDynamoTags.scala
@@ -1,0 +1,62 @@
+package uk.ac.wellcome.platform.archive.bagverifier.fixity.azure
+
+import com.amazonaws.services.dynamodbv2.AmazonDynamoDB
+import com.azure.storage.blob.BlobServiceClient
+import org.scanamo.auto._
+import org.scanamo.syntax._
+import org.scanamo.{Scanamo, Table}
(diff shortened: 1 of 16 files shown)


bag_verifier/src/main/scala/uk/ac/wellcome/platform/archive/bagverifier/fixity/azure/AzureDynamoTags.scala
```diff
@@ -0,0 +1,84 @@
...
+    tags: Map[String, String]
+  ): Either[WriteError, Map[String, String]] =
+    if (exists(location)) {
+      val ops = if (tags.isEmpty) {
+        table.delete('id -> location.toString())
+      } else {
+        table.put(DynamoTagsEntry(id = location.toString(), tags = tags))
+      }
+
+      scanamo.exec(ops) match {
```
alicefuzier, line 48: I know it's a draft, but why do we use scanamo direclty instead of the store interface?

alexwlchan: The Dynamo stores we've already got assume an ID and a version (even if the thing isn’t versioned); we don’t enforce versioning constraints on tags in S3.

Also I’m not entirely convinced those stores would handle tags correctly – in particular, if you write `CacheEntry(id = "1", tags = Map.empty)`, what actually gets written to DynamoDB is `"id" -> "1", "tags" -> null`, and then Scanamo has issues serialising it back out again.

alicefuzier (approved):
🎉

PR #731, opened 2020-09-10
refactor BagVerifier to have dependencies injected rather than instantiated internally

alicefuzier:


bag_verifier/src/main/scala/uk/ac/wellcome/platform/archive/bagverifier/services/s3/S3BagVerifier.scala
```diff
@@ -71,16 +62,62 @@ class S3StandaloneBagVerifier(val primaryBucket: String)(
...
       S3ObjectLocationPrefix
     ]
     with S3BagVerifier[
       ReplicatedBagVerifyContext[S3ObjectLocationPrefix]
-    ] {
+    ]
 
-  override val replicaStreamStore: StreamStore[S3ObjectLocation] =
-    new S3StreamStore()
+object S3BagVerifier {
```
alexwlchan, line 85: Minor: I would have made objects for each of S3StandaloneBagVerifier and S3ReplicatedBagVerifier, each with a single apply() method, but I'm not sure it makes a difference.

alexwlchan (approved):

PR #734, opened 2020-09-14
More dependency injection extraction

alicefuzier:


bag_verifier/src/main/scala/uk/ac/wellcome/platform/archive/bagverifier/services/BagVerifier.scala
```diff
@@ -59,6 +65,7 @@ trait BagVerifier[BagContext <: BagVerifyContext[BagPrefix], BagLocation <: Loca
     bag: Bag
   ): Either[BagVerifierError, Unit]
 
+  def isRetriable(error: BagVerifierError): Boolean
```
alexwlchan, line 68: nit: both “retriable” and “retryable” are correct spellings; elsewhere in the scala-libs we picked “retryable”. Could you change to be consistent?

And also add a newline to separate this from the comment.

bag_verifier/src/main/scala/uk/ac/wellcome/platform/archive/bagverifier/services/azure/AzureReplicatedBagVerifier.scala
```diff
@@ -37,6 +41,17 @@ class AzureReplicatedBagVerifier(
...
+  override def isRetriable(error: BagVerifierError): Boolean = error match {
+    // When reading from Azure, sometimes we see errors like
+    // "StoreReadError(reactor.core.Exceptions$ReactiveException: java.util.concurrent.TimeoutException: Did not
+    // observe any item or terminal signal within 60000ms in 'map' (and no fallback has been configured))"
+    // These are retriable errors so let's mark them as such
+    case BagVerifierError(e: RuntimeException, _)
+        if e.getMessage.contains("TimeoutException") =>
+      true
+    case _ => false
+  }
```
alexwlchan, line 54: Minor: since the second part of a BagVerifierError is always going to be a user-facing message, maybe match over `error.e` instead? i.e.

```scala
error.e match {
  case e: RuntimeException if e.getMessage.contains("TimeoutException") => true
  case _ => false
}
```

bag_verifier/src/main/scala/uk/ac/wellcome/platform/archive/bagverifier/services/s3/S3BagVerifier.scala
```diff
@@ -36,6 +36,8 @@ trait S3BagVerifier[B <: BagVerifyContext[S3ObjectLocationPrefix]]
     location: S3ObjectLocation
   ): String =
     location.key.replace(root.keyPrefix, "")
+
+  override def isRetriable(error: BagVerifierError): Boolean = false
```
alexwlchan, line 40: Maybe a quick comment about why there are no retryable S3 errors? (Because we've got that retrying logic elsewhere, presumably?)

alexwlchan (commented):
tl;dr: Adding retry logic in the FixityChecker is probably the wrong place, sorry for misleading you.

I started looking at this, and when I got to the`S3FixityChecker { def isRetriableError() = true }` I had a pause. Why is S3 able to do all its retrying elsewhere, but not Azure?

In the S3 error handler, it can see 500 errors from S3 and retry them: https://github.com/wellcomecollection/scala-libs/blob/master/storage/src/main/scala/uk/ac/wellcome/storage/s3/S3Errors.scala

Azure, on the other hand, is unable to do so: https://github.com/wellcomecollection/scala-libs/blob/master/storage/src/main/scala/uk/ac/wellcome/storage/azure/AzureStorageErrors.scala

And so when the Azure ranged reader goes to read a chunk of bytes from Azure, this exception bubbles up, and it doesn't get retried because the retry logic sees the exception as terminal.

In the LargeStreamReader (https://github.com/wellcomecollection/storage-service/blob/f7629d75b7679b75618579efb6b61e6ff913ab27/common/src/main/scala/uk/ac/wellcome/platform/archive/common/storage/services/LargeStreamReader.scala#L67-L89), the RetryOps will only retry the underlying getBytes call if the Left it gets back is a retryable error (https://github.com/wellcomecollection/scala-libs/blob/master/storage/src/main/scala/uk/ac/wellcome/storage/RetryOps.scala#L35-L49).

We should probably look at putting this "is it a timeout" in that AzureStorageErrors block instead, and testing it there, then letting it bubble up to all the other places. Sorry for misleading you.

The other changes to make FixityChecker composable look good at a glance, but we'll need to disentangle the retrying logic.

alicefuzier:
I don't think I understand your comment about the S3FixityChecker? I don't think I put the retry logic there unless I did something weird yesterday evening. The main changes in the FixityCheckers were meant to be around dependency injection cause I forgot them in the other PR but I was trying to write a test that was as close as possible to the real case, so with a failing streamReader used in the fixityChecker

alexwlchan:
Sorry, not the fixity checker, the S3 bag verifier. It was looking at this line:

```scala
override def isRetriable(error: BagVerifierError): Boolean = false
```

which made me wonder “hmm, why doesn’t the S3 bag verifier do any retrying 🤔” that made me realise it’s because we catch retryable S3 errors elsewhere.

alicefuzier:
ah ok makes sense. I had found where the error gets thrown but I had missed how the retry logic works. I'll change to use AzureStorageErrors

alexwlchan (approved):

PR #736, opened 2020-09-17
Implement a multi-size finder for S3

alexwlchan:
Epistemic status: moderately experimental. I'm not sure if this is a worthwhile tradeoff of code complexity/register performance, and I'd be interested in other opinions.

In the bag register, we often need to find lots of sizes at once – if the objects in a bag are all new, and don't have sizes listed in fetch.txt. Currently we do this by making an individual HeadObject call for each object. This is slow – registering a bag with ~37k objects takes 10 minutes, because we're making ~37k individual API calls.

We can get the size of 1000 objects at once by calling ListObjectsV2, which should make registration faster and more reliable. Call it when you're asked for the size of the first object, then return cached results for the next object along.

The comments in the code should explain how it works, hopefully.


alexwlchan:
Having written this and proven that the idea works, I'm now inclined to close this PR unless anybody else feels strongly.

Bag register reliability with large bags hasn't been a big issue in the past, and the speed of the pipeline is dominated by replication/verification.

Useful proof-of-concept, not sure it's worth deploying.

PR #739, opened 2020-09-18
If the ingest worker tries to update a non-existent ingest, retry

alexwlchan:
Closes https://github.com/wellcomecollection/platform/issues/4781

This introduces a new status code in the ingest tracker API. If you do a `PATCH /ingest/{id}` and that ingest doesn't exist yet, you get an HTTP 404 Not Found instead of HTTP 409 Conflict. This can occur if there's a DynamoDB consistency issue that means the DynamoTracker can't find the ingest.

The tracker client turns this status code into a new error type, and the ingest worker treats this error as non-deterministic.

Plus a couple of other changes to make this code easier to debug if we have other issues in future.

### DIFF
--- a/ingests/ingests_tracker/src/main/scala/uk/ac/wellcome/platform/storage/ingests_tracker/tracker/IngestError.scala
+++ b/ingests/ingests_tracker/src/main/scala/uk/ac/wellcome/platform/storage/ingests_tracker/tracker/IngestError.scala
@@ -4,6 +4,7 @@ import uk.ac.wellcome.platform.archive.common.bagit.models.BagVersion
 import uk.ac.wellcome.platform.archive.common.ingests.models.{Callback, Ingest}
 import uk.ac.wellcome.storage.{
   NotFoundError,
+  StorageError,
   UpdateNoSourceError,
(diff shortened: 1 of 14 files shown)


ingests/ingests_tracker/src/main/scala/uk/ac/wellcome/platform/storage/ingests_tracker/tracker/IngestError.scala
```diff
@@ -37,9 +38,9 @@ case class MismatchedVersionUpdateError(
 
 case class NoCallbackOnIngestError() extends IngestStoreError
 
-case class IngestStoreUnexpectedError(e: Throwable) extends IngestStoreError {
-
+case class IngestStoreUnexpectedError(storageError: StorageError)
+    extends IngestStoreError {
   override def toString: String = {
-    s"IngestStoreUnexpectedError: ${e.toString}"
```
alexwlchan, line 43: Discarding the outer wrapper meant the Throwable we got was sometimes `null`, which is unhelpful.

alicefuzier (approved):
🕵️

PR #750, opened 2020-09-29
Delete backfill infra

alicefuzier:
Resolves https://github.com/wellcomecollection/platform/issues/4797


terraform/modules/stack/iam_role_policy.tf
```diff
@@ -188,29 +188,6 @@ resource "aws_iam_role_policy" "allow_azure_verifier_access_cache" {
...
-  role   = module.bag_register_backfill.task_role_name
-  policy = data.aws_iam_policy_document.replica_primary_readonly.json
-}
-
-resource "aws_iam_role_policy" "bag_register_backfill_metrics" {
-  role   = module.bag_register_backfill.task_role_name
-  policy = data.aws_iam_policy_document.cloudwatch_putmetrics.json
-}
-
 resource "aws_iam_role_policy" "azure_replicator_replica_primary_readonly" {
```
alexwlchan, line 191: I think we can delete this policy too – the Azure replicator should be reading from the unpacker bucket, not the primary storage.

terraform/stack_prod/main.tf
```diff
@@ -13,7 +13,7 @@ module "stack_prod" {
   domain_name      = local.domain_name
   cert_domain_name = local.cert_domain_name
 
-  min_capacity = 0
+  min_capacity = 1
```
alexwlchan, line 16:
```suggestion
  min_capacity = 0
```

terraform/stack_staging/main.tf
```diff
@@ -13,7 +13,7 @@ module "stack_staging" {
   domain_name      = local.staging_domain_name
   cert_domain_name = local.cert_domain_name
 
-  min_capacity = 0
+  min_capacity = 1
```
alexwlchan, line 16:
```suggestion
  min_capacity = 0
```

We let the service scale down when it's not in use.

alexwlchan (approved):
🔥🔥🔥

Looks good, modulo a couple of suggestions.

PR #751, opened 2020-09-30
Rewrite the daily reporter to be more useful

alexwlchan:
Closes https://github.com/wellcomecollection/platform/issues/4769

New features:

* Previously the daily reporter would only report on ingests that were updated in the last 2 days. It now additionally looks for ingests from any time which are in the "processing" or "accepted" states, which might indicate a stalled ingest.

* Stalled ingests are now flagged in the daily Slack message.

* It writes a colourful HTML report of the last two days of storage service activity, which is written to a public S3 bucket configured to do static website hosting: https://wellcomecollection-storage-daily-reporting.s3-eu-west-1.amazonaws.com/2020-09-30.html

### DIFF
--- a/monitoring/daily_reporter/requirements.in
+++ b/monitoring/daily_reporter/requirements.in
@@ -1,2 +1,3 @@
 boto3
 httpx
+jinja2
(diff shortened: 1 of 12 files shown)


alexwlchan:
Merging so I can get the Terraform changes this has.

PR #754, opened 2020-10-01
Send a test bag in stage from BuildKite

alexwlchan:
For https://github.com/wellcomecollection/platform/issues/4819

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -1,55 +1,61 @@
 steps:
-  - command: .buildkite/scripts/run_autoformat.py
-    label: "autoformat"
-  - wait
-  - command: .buildkite/scripts/run_job.py common
-    label: "common"
-  - command: .buildkite/scripts/run_job.py display
-    label: "display"
-  - command: .buildkite/scripts/run_job.py indexer_common
-    label: "indexer_common"
-  - wait
-  - command: .buildkite/scripts/run_job.py bags_api
-    label: "bags API"
-  - command: .buildkite/scripts/run_job.py bag_register
-    label: "bag register"
-  - command: .buildkite/scripts/run_job.py bag_replicator
-    label: "bag replicator"
-  - command: .buildkite/scripts/run_job.py bag_root_finder
-    label: "bag root finder"
-  - command: .buildkite/scripts/run_job.py bag_tagger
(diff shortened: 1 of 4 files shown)


.buildkite/pipeline.yml
```diff
@@ -53,3 +53,16 @@ steps:
           workdir: /repo
           mount-ssh-agent: true
           command: ["--confirm", "release-deploy", "--from-label", "latest", "--environment-id", "stage", "--description", $BUILDKITE_BUILD_URL]
+  - wait
+  - label: send a test bag in stage
+    if: build.branch == "master"
+    plugins:
+      - docker#v3.5.0:
+          image: "amazon/aws-cli"
+          mount-ssh-agent: true
```
kenoir, line 62: I'm not certain you need this.

terraform/monitoring/lambda_end_to_end_bag_test.tf
```diff
@@ -37,3 +37,31 @@ module "end_to_end_bag_tester_prod" {
 
   tags = local.default_tags
 }
+
+# Allow the CI agent running in BuildKite to trigger the Lambda after
+# deploying new images to staging.
+# See https://github.com/wellcomecollection/platform/issues/4819
+resource "aws_lambda_permission" "allow_platform_to_trigger_lambda" {
```
kenoir, line 44: Does it make sense to describe these CI permissions where the CI role is created in platform-infra (I think there are other specific permissions there).

kenoir (approved):

.buildkite/pipeline.yml
alexwlchan: Ah, possibly not. Do AWS credentials get exposed inside the container then?

kenoir: The containers can access the host credentials via the network endpoint available on EC2 instances.

kenoir: So basically it's transparent to anything that uses an AWS SDK.

terraform/monitoring/lambda_end_to_end_bag_test.tf
alexwlchan: I thought about it; it's not really obvious to me that one makes more sense than the other.

I picked here just because it means one PR rather than two.

PR #760, opened 2020-10-08
Use an emoji to highlight daily reporter state

alexwlchan:
Stealing an idea from @kenoir in the snapshot reporter: if all is well, it prints ✅. If something needs attention, it prints ⁉️


kenoir (approved):
✅

PR #762, opened 2020-10-13
Archivematica Bag Migration

kenoir:
Part of https://github.com/wellcomecollection/platform/issues/4799

Adds a script to get bag data from the reporting cluster, needed to create a new set of bags.


scripts/archivematica_bags_migration.py
```diff
@@ -0,0 +1,212 @@
...
+    assumed_role_object = sts_client.assume_role(
+        RoleArn=role_arn, RoleSessionName="AssumeRoleSession1"
+    )
+    credentials = assumed_role_object["Credentials"]
+    return boto3.client(
+        resource,
+        aws_access_key_id=credentials["AccessKeyId"],
+        aws_secret_access_key=credentials["SecretAccessKey"],
+        aws_session_token=credentials["SessionToken"],
+    )
```
alexwlchan, line 36: Note: our pattern (at your suggestion IIRC) is to put one-off scripts like this in the `migrations` folder. Once you're there, stuff like this is available to import from `common.py`.

scripts/archivematica_bags_migration.py
```diff
@@ -0,0 +1,212 @@
...
+                save_path = f"{working_folder}/{filename}"
+                s3_client.download_file(bucket, location, save_path)
+
+        def _compress_bag():
+            shutil.make_archive(working_folder, "gztar", working_folder)
+            shutil.rmtree(working_folder, ignore_errors=True)
+
+        _write_fetch_file()
+        _get_bag_files()
+        # Get Archivematica UUID ???
```
alexwlchan, line 164: If you look in an Archivematica bag, there's a METS file in the top level called `METS.[UUID].xml`. That UUID is the one you want.

scripts/migrations/2020-10-archivematica_bags_migration.py
```diff
@@ -0,0 +1,246 @@
...
+        def _check_expected_prefixes_exist(file_locations):
+            for prefix in prefix_patterns:
+                found_match = False
+
+                for location in file_locations:
+                    if prefix in location:
+                        found_match = True
+
+                if not found_match:
+                    raise RuntimeError(f"Missing any files matching prefix: {prefix}")
```
alexwlchan, line 100: Python has a convenience method you might find helpful

https://docs.python.org/3/library/functions.html?highlight=any#any:

> Return True if any element of the iterable is true. If the iterable is empty, return False.

Consider:

```suggestion
            for prefix in prefix_patterns:
                if not any(prefix in location for location in file_locations):
                    raise RuntimeError(f"Missing any files matching prefix: {prefix}")
```

scripts/migrations/2020-10-archivematica_bags_migration.py
```diff
@@ -0,0 +1,246 @@
...
+            file_locations = []
+            for prefix in prefix_patterns:
+                file_locations = file_locations + _filter_bag_files(prefix)
+
+            _check_expected_prefixes_exist(file_locations)
+
+            for location in file_locations:
+                filename = location.split("/")[-1]
+                save_path = f"{working_folder}/{filename}"
+                storage_s3_client.download_file(bucket, location, save_path)
```
alexwlchan, line 112: Couple of suggestions:

* I would call this _get_bag_manifest_files to be really explicit that you mean things like fetch.txt and the tag manifest, not payload files.
* You can already get a list of these files from the `"tagManifest"` entry on a JSON storage manifest from the storage service.

scripts/migrations/2020-10-archivematica_bags_migration.py
```diff
@@ -0,0 +1,246 @@
...
+            shutil.rmtree(working_folder, ignore_errors=True)
+
+        def _generate_checksum(file_location):
+            sha256_hash = hashlib.sha256()
+
+            with open(file_location, "rb") as f:
+                for byte_block in iter(lambda: f.read(4096), b""):
+                    sha256_hash.update(byte_block)
+
+                return sha256_hash.hexdigest()
```
alexwlchan, line 125: Nice, could be a standalone function which would make the bag_creator function easier to read.

scripts/migrations/2020-10-archivematica_bags_migration.py
```diff
@@ -0,0 +1,246 @@
...
+            merged_checksums = dict(existing_checksums.items() | new_checksums.items())
+
+            with open(f"{working_folder}/{tagmanifest_name}", "w") as fp:
+                for checksum, filename in merged_checksums.items():
+                    fp.write(f"{filename} {checksum}\n")
+
+        def _get_archivematica_uuid():
+            files = _filter_bag_files("data/METS.")
+
+            assert len(files) == 1
```
alexwlchan, line 164: minor: Python will just give you a generic "AssertionFailed" error if this goes wrong. I like to include the value I was just testing, so I've got a chance of debugging it:

```suggestion
            assert len(files) == 1, files
```

(Some test frameworks will give you nicer assertions, but not vanilla Python.)

scripts/migrations/2020-10-archivematica_bags_migration.py
```diff
@@ -0,0 +1,246 @@
...
+            mets_file_with_id = files[0]
+
+            archivematica_uuid = mets_file_with_id.split("/METS.")[-1].split(".xml")[0]
+
+            assert archivematica_uuid
```
alexwlchan, line 169: Checking the UUID is non-empty – nice. Even nicer: how about using the uuid library to check this is actually a UUID?

scripts/migrations/2020-10-archivematica_bags_migration.py
```diff
@@ -0,0 +1,246 @@
...
+            return archivematica_uuid
+
+        def _update_bag_info():
+            archivematica_uuid = _get_archivematica_uuid()
+            with open(f"{working_folder}/bag-info.txt", "a") as fp:
+                fp.write(f"Internal-Sender-Identifier: {archivematica_uuid}")
```
alexwlchan, line 175: Don't forget the trailing newline!

```suggestion
                fp.write(f"Internal-Sender-Identifier: {archivematica_uuid}\n")
```

scripts/migrations/2020-10-archivematica_bags_migration.py
```diff
@@ -0,0 +1,246 @@
...
+            }
+
+        def _load_existing_checksums():
+            tag_manifest = {}
+
+            with open(f"{working_folder}/{tagmanifest_name}") as fp:
+                for _, line in enumerate(fp):
+                    split_manifest_line = line.split(" ")
+                    checksum = split_manifest_line[0].strip()
+                    filename = split_manifest_line[1].strip()
```
alexwlchan, line 142: I'd suggest using tuple unpacking here:

```suggestion
                    checksum, filename = line.strip().split()
```

That way, Python will warn you if there isn't enough text in the line (or conversely, too much!).

scripts/migrations/2020-10-archivematica_bags_migration.py
```diff
@@ -0,0 +1,246 @@
...
+
+                    assert checksum
+                    assert filename
+
+                    tag_manifest[filename] = checksum
```
alexwlchan, line 147: I would throw in an assertion that `filename` isn't already in this dictionary, which would surely indicate something very weird is up.

scripts/migrations/2020-10-archivematica_bags_migration.py
```diff
@@ -0,0 +1,246 @@
...
+            return {
+                filename: _generate_checksum(f"{working_folder}/{filename}")
+                for filename in files_in_need_of_update
+            }
+
+        def _load_existing_checksums():
+            tag_manifest = {}
+
+            with open(f"{working_folder}/{tagmanifest_name}") as fp:
+                for _, line in enumerate(fp):
```
alexwlchan, line 139: minor: don't think enumerate adds anything?

```suggestion
                for line in fp:
```

scripts/migrations/2020-10-archivematica_bags_migration.py
```diff
@@ -0,0 +1,246 @@
...
+
+            _check_expected_prefixes_exist(file_locations)
+
+            for location in file_locations:
+                filename = location.split("/")[-1]
+                save_path = f"{working_folder}/{filename}"
+                storage_s3_client.download_file(bucket, location, save_path)
+
+        def _compress_bag():
+            shutil.make_archive(working_folder, "gztar", working_folder)
```
alexwlchan, line 115: This method returns the name of the archive it generates, which is more robust than hard-coding it as `archive_name` above.

scripts/migrations/2020-10-archivematica_bags_migration.py
```diff
@@ -0,0 +1,246 @@
...
+            )
+
+        def _request_ingest():
+            response = storage_client.create_s3_ingest(
+                space=space,
+                external_identifier=external_identifier,
+                s3_bucket=s3_upload_bucket,
+                s3_key=f"{s3_upload_prefix}/{archive_name}",
+                ingest_type="update",
+            )
```
alexwlchan, line 191: This response should include the ingest ID in the header – how about printing it, so it's easy to find?

scripts/migrations/2020-10-archivematica_bags_migration.py
```diff
@@ -0,0 +1,246 @@
...
+
+    storage_s3_client = get_aws_client(resource="s3", role_arn=storage_role_arn)
+
+    elastic_client = get_elastic_client(
+        role_arn=storage_role_arn, elastic_secret_id=elastic_secret_id
+    )
+
+    storage_client = get_storage_client(api_url=api_url)
+
+    # TODO: update query to include born-digital-accessions
```
alexwlchan, line 238: And to ignore bags that already have an Internal-Sender-Identifier field, so you don't send multiple updates for the same bag.

scripts/migrations/common.py
```diff
@@ -101,3 +106,41 @@ def get_aws_client(resource, *, role_arn):
         aws_secret_access_key=credentials["SecretAccessKey"],
         aws_session_token=credentials["SessionToken"],
     )
+
+
+def get_storage_client(api_url):
```
alexwlchan, line 111: This can be cached as well – the storage service class has its own logic for refreshing tokens. You don't need a new instance unless the on-disk credentials have changed.

alexwlchan (commented):
Bunch of comments inline, but mostly looks pretty good. My big concern is the amount of stuff happening in `bag_creator()`, a lot of which is fairly generic logic not specific to the process of bag creation. Pulling it out into separate functions would make it all easier to read.

scripts/migrations/2020-10-archivematica_bags_migration.py
kenoir: Will log it!

scripts/migrations/2020-10-archivematica_bags_migration.py
```diff
@@ -0,0 +1,349 @@
...
+        bucket = storage_manifest["location"]["bucket"]
+        path = storage_manifest["location"]["path"]
+        files = storage_manifest["manifest"]["files"]
+        provider = storage_manifest["location"]["provider"]["id"]
+        tagmanifest_files = storage_manifest["tagManifest"]["files"]
+
+        assert provider == "amazon-s3"
+
+        working_id = id.replace("/", "_")
+        working_folder = f"{self.target_folder}/{working_id}"
```
alexwlchan, line 210: minor: prefer `os.path.join(…)` to combining paths directly. Not an issue here, but bit more portable in general.

scripts/migrations/2020-10-archivematica_bags_migration.py
```diff
@@ -0,0 +1,349 @@
...
+        bucket, path, version, working_folder, tagmanifest_files
+    ):
+        file_locations = [
+            f"{path}/{version}/{file['name']}" for file in tagmanifest_files
+        ]
+
+        for location in file_locations:
+            filename = location.split("/")[-1]
+            save_path = f"{working_folder}/{filename}"
+            storage_s3_client.download_file(bucket, location, save_path)
```
alexwlchan, line 148: minor: prefer `os.path.basename(…)` instead of manipulating paths directly, for portability.

Also, you already have the filename as `file['name']`.

scripts/migrations/2020-10-archivematica_bags_migration.py
```diff
@@ -0,0 +1,349 @@
...
+        self.tagmanifest_name = "tagmanifest-sha256.txt"
+        self.s3_upload_prefix = "born-digital/archivematica-uuid-update"
+
+    @staticmethod
+    def _get_archivematica_uuid(bucket, path, version):
+        files = filter_s3_objects(
+            s3_client=storage_s3_client,
+            bucket=bucket,
+            prefix=f"{path}/{version}/data/METS.",
+        )
```
alexwlchan, line 91: You could save yourself a trip to S3 here – you have the payload files from the storage manifest. Something like:

```python
mets_files = [f for f in files if f.startswith("data/METS.") and f.endswith(".xml")]
```

would be faster.

scripts/migrations/2020-10-archivematica_bags_migration.py
```diff
@@ -0,0 +1,349 @@
...
+
+        assert len(files) == 1, files
+        mets_file_with_id = files[0]
+
+        archivematica_uuid = mets_file_with_id.split("/METS.")[-1].split(".xml")[0]
+
+        assert UUID(archivematica_uuid, version=4)
```
alexwlchan, line 98: 👍

scripts/migrations/2020-10-archivematica_bags_migration.py
```diff
@@ -0,0 +1,349 @@
...
+
+    def migrate(self, version, space, external_identifier):
+        storage_manifest = storage_client.get_bag(
+            space=space, external_identifier=external_identifier, version=version
+        )
+
+        id = storage_manifest["id"]
+        bucket = storage_manifest["location"]["bucket"]
+        path = storage_manifest["location"]["path"]
+        files = storage_manifest["manifest"]["files"]
```
alexwlchan, line 203: nit: I would call these `payload_files`.

scripts/migrations/2020-10-archivematica_bags_migration.py
```diff
@@ -0,0 +1,349 @@
...
+        )
+
+        did_append_uuid = self._append_archivematica_uuid(
+            working_folder, archivematica_uuid
+        )
+        if not did_append_uuid:
+            _log(
+                f"Internal-Sender-Identifier found in bag-info.txt: {archivematica_uuid}"
+            )
+            _log(f"Not migrating {id} (already migrated)")
```
alexwlchan, line 252: Small optimisation: once you've got the bag from the storage-service API, looks for `Internal-Sender-Identifier` in the API response. Saves you doing any downloads from S3 before you skip it here.

scripts/migrations/2020-10-archivematica_bags_migration.py
```diff
@@ -0,0 +1,349 @@
...
+def filter_s3_objects(s3_client, bucket, prefix):
+    response = s3_client.list_objects_v2(Bucket=bucket, Prefix=prefix)
+
+    if "Contents" in response:
+        return [content["Key"] for content in response["Contents"]]
+    else:
+        return []
+
+
+def load_space_separated_file(file_location, key_first=True):
```
alexwlchan, line 57: minor: took me a while to figure out what this function was for – FWIW, I think the term is "tag file".

scripts/migrations/2020-10-archivematica_bags_migration.py
```diff
@@ -0,0 +1,349 @@
...
+        )
+
+        if "Internal-Sender-Identifier:" in bag_info:
+            assert (
+                bag_info["Internal-Sender-Identifier:"] == archivematica_uuid
+            ), archivematica_uuid
+            return False
+        else:
+            with open(f"{working_folder}/bag-info.txt", "a") as fp:
+                fp.write(f"Internal-Sender-Identifier: {archivematica_uuid}\n")
```
alexwlchan, line 163: What if the existing bag-info.txt doesn't end with a newline?

alexwlchan (approved):
A couple of suggestions for improvements to make if this was Python we expected to maintain long term, but fine for doing a one-off migration.

Don't forget to do the accessions as well! And you can save yourself some work by filtering out already-migrated bags sooner.

scripts/migrations/2020-10-archivematica_bags_migration.py
```diff
@@ -0,0 +1,349 @@
...
+
+
+if __name__ == "__main__":
+    storage_role_arn = "arn:aws:iam::975596993436:role/storage-developer"
+    workflow_role_arn = "arn:aws:iam::299497370133:role/workflow-developer"
+
+    elastic_secret_id = "archivematica_bags_migration/credentials"
+    index = "storage_bags"
+
+    environment_id = "prod"
```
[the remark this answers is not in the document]
kenoir: I have this locally 👍 will push.

scripts/migrations/2020-10-archivematica_bags_migration.py
```diff
@@ -0,0 +1,349 @@
...
+
+    storage_s3_client = get_aws_client(resource="s3", role_arn=storage_role_arn)
+
+    elastic_client = get_elastic_client(
+        role_arn=storage_role_arn, elastic_secret_id=elastic_secret_id
+    )
+
+    storage_client = get_storage_client(api_url=api_url)
+
+    elastic_query = {"query": {"prefix": {"space": {"value": "born-digital"}}}}
```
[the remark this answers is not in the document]
kenoir: > Getting stuff from the born-digital-accessions space
`prefix` will find those also

> Filtering out bags that already have an I-S-I.
Not filtering them out, but also not migrating them (see https://github.com/wellcomecollection/storage-service/pull/762/files#diff-a8c28d3bd1fd22a2f78607385291974d5124ac1da1954c2cc0440b6913c562e2R249).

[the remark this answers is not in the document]
alexwlchan: Yeah, I saw that later. I would do it ASAP to avoid doing unnecessary work, but not a big deal for a one-off migration.

PR #769, opened 2020-10-22
Miro Storage Service migration scripts

kenoir:
Part of https://github.com/wellcomecollection/platform/issues/4809


scripts/migrations/2020-10-miro-migration/common.py
```diff
@@ -0,0 +1,151 @@
...
+    )
+
+    oauth_creds = json.load(open(creds_path))
+
+    return wellcome_storage_service.RequestsOAuthStorageServiceClient(
+        api_url=api_url,
+        client_id=oauth_creds["client_id"],
+        client_secret=oauth_creds["client_secret"],
+        token_url=oauth_creds["token_url"],
+    )
```
alexwlchan, line 124: FYI, you should be able to replace this with

```python
return wellcome_storage_service.RequestsOAuthStorageServiceClient.from_path(api_url=api_url)
```

scripts/migrations/2020-10-miro-migration/main.py
```diff
@@ -0,0 +1,150 @@
...
+    return query_results["hits"]["hits"]
+
+
+@click.command()
+@click.pass_context
+def create_files_index(ctx):
+    local_elastic_client = ctx.obj["local_elastic_client"]
+    reporting_elastic_client = ctx.obj["reporting_elastic_client"]
+
+    expected_file_count = 223_528
```
alexwlchan, line 84: Where's this count from?

scripts/migrations/2020-10-miro-migration/main.py
```diff
@@ -0,0 +1,150 @@
...
+    local_file_index = "files"
+
+    click.echo(f"Recreating files index ({local_file_index})")
+
+    local_elastic_client.indices.delete(index=local_file_index, ignore=[400, 404])
+
+    create_index(elastic_client=local_elastic_client, index_name=local_file_index)
```
alexwlchan, line 91: Does making this a function add anything?

scripts/migrations/2020-10-miro-migration/main.py
```diff
@@ -0,0 +1,150 @@
...
+
+def s3_miro_objects(s3_client):
+    for prefix_path in S3_MIRO_PREFIX_PATHS:
+        filtered_path_prefix = f"{S3_MIRO_IMAGES_PATH}/{prefix_path}/"
+        filtered_s3_objects = filter_s3_objects(
+            s3_client=s3_client, bucket=S3_MIRO_BUCKET, prefix=filtered_path_prefix
+        )
+
+        for s3_object in filtered_s3_objects:
+            truncated_path = s3_object["Key"].replace(filtered_path_prefix, "")
```
alexwlchan, line 57: I think you'd get the same effect with:

```python
truncated_path = os.path.relpath(s3_object["Key"], relpath=filtered_path_prefix)
chunk = os.path.basename(truncated_path)
```

alexwlchan (commented):
Looks good at a quick glance; will run properly tomorrow.

scripts/migrations/2020-10-miro-migration/transfer_packager.py
```diff
@@ -6,22 +6,73 @@
...
+        s3_client.download_file(
+            Bucket=s3_bucket,
+            Key=s3_key,
+            Filename=file_location
+        )
+
+    # Check if we already have this file
+    if os.path.isfile(file_location):
+        # If the length doesn't match overwrite
+        if not os.path.getsize(file_location) == s3_content_length:
```
jamieparkinson, line 37: nit: personally find the nested `if if else` a bit tricky to read, maybe prefer
```python
file_exists = os.path.isfile(file_location)
file_size_matches = os.path.getsize(file_location) == s3_content_length
if file_exists and not file_size_matches:
  ...
```

jamieparkinson: I am wrong

scripts/migrations/2020-10-miro-migration/chunk_transfer.py
```diff
@@ -0,0 +1,46 @@
+import elasticsearch
+
+from chunks import Chunk
+from common import get_aws_client
+from elastic_helpers import get_local_elastic_client
+from transfer_packager import create_transfer_package, upload_transfer_package
+
+STORAGE_ROLE_ARN = "arn:aws:iam::975596993436:role/storage-developer"
```
alexwlchan, line 8: We should be using the read-only role here, I think?

scripts/migrations/2020-10-miro-migration/transfer_packager.py
```diff
@@ -0,0 +1,166 @@
...
+
+    assert (
+        local_content_length == expected_content_length
+    ), f"Content length mismatch: {file_location}!"
+
+
+def _get_s3_content_length(s3_client, s3_bucket, s3_key):
+    s3_head_object_response = s3_client.head_object(Bucket=s3_bucket, Key=s3_key)
+
+    return s3_head_object_response["ContentLength"]
```
alexwlchan, line 26: This is the sort of thing you could push into `s3.py` to simplify this file – throw in a docstring and this is a nice, reusable function.

scripts/migrations/2020-10-miro-migration/transfer_packager.py
```diff
@@ -0,0 +1,166 @@
...
+    )
+
+    def _download():
+        s3_client.download_file(Bucket=s3_bucket, Key=s3_key, Filename=file_location)
+
+    # Check if we already have this file
+    if os.path.isfile(file_location):
+        # If the length doesn't match overwrite
+        if not os.path.getsize(file_location) == s3_content_length:
+            _download()
```
alexwlchan, line 44: I assume this is the bit we want to change to support multiple filenames? Let's discuss exactly how we download files.

scripts/migrations/2020-10-miro-migration/transfer_packager.py
```diff
@@ -0,0 +1,166 @@
...
+def _assert_local_content_length(file_location, expected_content_length):
+    # Assert we have the correct number of bytes
+    local_content_length = os.path.getsize(file_location)
+
+    # Check that content length is non-zero
+    assert local_content_length > 0
+
+    assert (
+        local_content_length == expected_content_length
+    ), f"Content length mismatch: {file_location}!"
```
alexwlchan, line 20: Notes:

* I would call this `check` rather than `assert`. It does an assertion, but the purpose of this function is to check the content length matches than in S3.
* The first comment should be the docstring, not a comment.
* I would log the expected/actual content lengths in the assertion.

scripts/migrations/2020-10-miro-migration/transfer_packager.py
```diff
@@ -0,0 +1,166 @@
...
+
+                # Schedule the next set of futures.  We don't want more than N futures
+                # in the pool at a time, to keep memory consumption down.
+                for task in itertools.islice(s3_key_list_iter, len(done)):
+                    fut = executor.submit(_get_s3_object, task)
+                    futures[fut] = task
+
+        assert len(os.listdir(target_folder)) == len(
+            s3_key_list
+        ), f"Unexpected file count in {target_folder}!"
```
alexwlchan, line 94: Nice – I was about to suggest something like this. I suspect the S3 downloader may leave temporary files hanging around if it's interrupted.

scripts/migrations/2020-10-miro-migration/chunks.py
```diff
@@ -0,0 +1,68 @@
...
+
+@attr.s
+class Chunk:
+    miro_shard = attr.ib()
+    destination = attr.ib()
+    s3_keys = attr.ib(default=[])
+
+    def merge_chunk(self, chunk):
+        assert chunk.chunk_id() == self.chunk_id()
+        self.s3_keys = self.s3_keys + chunk.s3_keys
```
alexwlchan, line 24: Nit: a common convention is to use `other` for this sort of thing in Python.

scripts/migrations/2020-10-miro-migration/chunks.py
```diff
@@ -0,0 +1,68 @@
...
+
+from decisions import Decision
+from elastic_helpers import get_local_elastic_client
+
+
+@attr.s
+class Chunk:
+    miro_shard = attr.ib()
+    destination = attr.ib()
+    s3_keys = attr.ib(default=[])
```
alexwlchan, line 20: Beware lists as defaults! Unlike Scala, Python evaluates default arguments *once* and then reuses the same instance everywhere. This can lead to unexpected behaviour if those defaults are mutable e.g.

```pycon
>>> def appender(values=[]):
...     values.append(1)
...     print(values)
...
>>> appender([1, 2, 3])
[1, 2, 3, 1]
>>> appender([1, 2, 4])
[1, 2, 4, 1]
>>> appender()
[1]
>>> appender()
[1, 1]
```

See https://www.attrs.org/en/stable/init.html#defaults

scripts/migrations/2020-10-miro-migration/chunks.py
```diff
@@ -0,0 +1,68 @@
...
+                destination=destination,
+                s3_keys=[decision.s3_key],
+            )
+
+            chunk_id = new_chunk.chunk_id()
+
+            if chunk_id in groups:
+                groups[chunk_id].merge_chunk(new_chunk)
+            else:
+                groups[chunk_id] = new_chunk
```
alexwlchan, line 64: I would use a defaultdict and do this without the custom class. Something like:

```python
# Dict (shard, destination) -> set(s3 keys)
groups = collections.defaultdict(set)

for result in chunkable_decisions:
    decision = Decision(**result["_source"])
    for destination in decision.destinations:
        groups[(decision.miro_shard, destination)].add(decision.s3_key)

return groups
```

kenoir: I tried this with a defaultdict but prefer enforcing the shape of a "chunk" with the custom class (it will be used outside the module).

scripts/migrations/2020-10-miro-migration/transfer_packager.py
```diff
@@ -0,0 +1,195 @@
...
+            os.remove(file_location)
+
+    transfer_package.s3_location = s3_key
+
+    return transfer_package
+
+
+def create_transfer_package(s3_client, group_name, s3_bucket, s3_key_list, prefix):
+    target_base_folder = "target"
+    folder_id = str(uuid.uuid4())
```
alexwlchan, line 174: I would throw at https://leancrew.com/all-this/2014/10/asciifying/ at the group name

scripts/migrations/2020-10-miro-migration/transfer_packager.py
```diff
@@ -0,0 +1,195 @@
...
+                pbar.update(len(done))
+
+                # Schedule the next set of futures.  We don't want more than N futures
+                # in the pool at a time, to keep memory consumption down.
+                for task in itertools.islice(s3_key_list_iter, len(done)):
+                    fut = executor.submit(_get_s3_object, task)
+                    futures[fut] = task
+
+        target_folder_count = sum(
+            [len(files) for r, d, files in os.walk(target_folder)]
```
alexwlchan, line 106: nit: Python convention is to use `_` for unused variables, i.e.
```suggestion
            [len(files) for _, _, files in os.walk(target_folder)]
```

scripts/migrations/2020-10-miro-migration/transfer_packager.py
```diff
@@ -0,0 +1,195 @@
...
+        )
+
+        assert target_folder_count == len(s3_key_list), (
+            f"Unexpected file count in {target_folder}: "
+            f"{target_folder_count} != {len(s3_key_list)}"
+        )
```
alexwlchan, line 112: Nice.

scripts/migrations/2020-10-miro-migration/transfer_packager.py
```diff
@@ -0,0 +1,195 @@
...
+
+
+def _compress_folder(target_folder, remove_folder=True):
+    archive_name = shutil.make_archive(target_folder, "zip", target_folder)
+    if remove_folder:
+        shutil.rmtree(target_folder, ignore_errors=True)
```
alexwlchan, line 118: minor: You only call `_compress_folder` once, and you don't override `remove_folder`. I would be explicit, remove the parameter, don't hide this behind a conditional.

scripts/migrations/2020-10-miro-migration/transfer_packager.py
```diff
@@ -0,0 +1,195 @@
...
+
+    def _download():
+        required_dir = os.path.dirname(file_location)
+        os.makedirs(required_dir, exist_ok=True)
+        s3_client.download_file(Bucket=s3_bucket, Key=s3_key, Filename=file_location)
+
+    # Check if we already have this file
+    if os.path.isfile(file_location):
+        # If the length doesn't match overwrite
+        if not os.path.getsize(file_location) == s3_content_length:
```
alexwlchan, line 55: nit: prefer `!=` to `not … ==`, i.e.,

```suggestion
        if os.path.getsize(file_location) != s3_content_length:
```

scripts/migrations/2020-10-miro-migration/s3.py
```diff
@@ -54,3 +53,12 @@ def get_s3_object(*, bucket, key):
...
         return open(out_path, "rb")
+
+
+def get_s3_content_length(s3_client, s3_bucket, s3_key):
+    """
+    Retrieves the content length of an object from S3.
+    """
+    s3_head_object_response = s3_client.head_object(Bucket=s3_bucket, Key=s3_key)
+
+    return s3_head_object_response["ContentLength"]
```
alexwlchan, line 64: minor: You can drop the "s3_" prefixes on most of the variables IMO, and I'd name this something like `get_s3_object_size`. That it's the ContentLength field in S3 is a bit of an implementation detail.

scripts/migrations/2020-10-miro-migration/transfer_packager.py
```diff
@@ -0,0 +1,195 @@
...
+    """
+    local_content_length = os.path.getsize(file_location)
+
+    assert local_content_length > 0, "Content length is zero: " f"{file_location}"
+
+    assert local_content_length == expected_content_length, (
+        "Content length mismatch "
+        f"({local_content_length} != {expected_content_length}): "
+        f"{file_location}"
+    )
```
alexwlchan, line 34: Ditto the S3 function above: `_check_file_size_matches()` makes more sense than talking about Content-Length.

scripts/migrations/2020-10-miro-migration/transfer_packager.py
kenoir: It's there as a convenience for me to see the folder if I need to ... but I guess it's as easy to just comment this line out if I need to.

kenoir:
Closing this PR , but keeping the branch for posterity.

PR #776, opened 2020-11-16
Expire non-current versions in our permanent storage buckets

alexwlchan:
Although we have versioning enabled, we never actually delete the old versions once they're gone!

I've applied this in staging and it looks fine (although we don't actually enable versioning on the staging buckets); I'll apply in prod once this has been reviewed.

### DIFF
--- a/terraform/modules/critical/s3_replica_glacier.tf
+++ b/terraform/modules/critical/s3_replica_glacier.tf
@@ -16,6 +16,25 @@ resource "aws_s3_bucket" "replica_glacier" {
     }
   }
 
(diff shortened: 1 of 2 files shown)


jamesgorrie (commented):
Just to have context as to why we might chose these times what would we use previous versions for?

alexwlchan:
> Just to have context as to why we might chose these times what would we use previous versions for?

There is some reasoning behind this, but it belongs in the code rather than a PR discussion. I'll push another commit in a bit.

kenoir (approved):

PR #777, opened 2020-11-16
Add a script for deleting bags

alexwlchan:
Checklist:

- [x] Make sure the user is really, really, really sure they want to do this
- [x] Record the deletion in a DynamoDB table
- [x] Squirrel away one final copy of the bag before deleting it
- [x] Delete the replicas in S3
- [x] Delete the replicas in Azure
- [ ] Delete the versioner/replica aggregator records
- [ ] Empty the reporting cluster
- [ ] Delete the storage manifest
- [ ] Delete the ingest

Closes https://github.com/wellcomecollection/platform/issues/4878

PR #783, opened 2020-11-18
Add more S3 utility functions for the "delete bags" script

alexwlchan:
Part of https://github.com/wellcomecollection/platform/issues/4878, spun out of https://github.com/wellcomecollection/storage-service/pull/777

### DIFF
--- a/scripts/helpers/iterators.py
+++ b/scripts/helpers/iterators.py
@@ -1,0 +1,19 @@
+import itertools
+
+
+def chunked_iterable(iterable, *, size):
+    """
+    Given an iterable, divide it into chunks of fixed size.
+
+    >>> chunked_iterable(range(10), size=3)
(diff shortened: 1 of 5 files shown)


scripts/helpers/iterators.py
```diff
@@ -0,0 +1,19 @@
+import itertools
+
+
+def chunked_iterable(iterable, *, size):
```
kenoir, line 4: This function is everywhere!

alexwlchan: It's so handy! 😜 

Which is kinda the point of starting to library-ise some of this stuff. These are utilities so often and rely upon, I want to spend less time copying them around, and more time trusting they work.

kenoir: They should put this one in core Python 😉

kenoir (approved):

PR #785, opened 2020-11-18
Add a script for deleting bags

alexwlchan:
Closes https://github.com/wellcomecollection/storage-service/pull/777, part of wellcomecollection/platform#4878

This is my third/fourth pass at this, and this one finally looks good. We've made it quite tricky to delete bags, so undoing those protections takes a while. 😅 

This is a reviewable unit – it does all the checking that yes, you really want to delete this bag, and then purges it from the reporting cluster. I'll add the rest in another PR.

### DIFF
--- a/scripts/ss_delete_bag.py
+++ b/scripts/ss_delete_bag.py
@@ -145,13 +145,14 @@ def _confirm_user_wants_to_delete_bag(
 
     It presents a prompt of the following form:
 
-          This is the bag you are about to delete:
-          Space:        testing
-          External ID:  test_bag
-          Version:      v132
-          Date created: Friday, 13 November 2020 @ 11:10 (2 days ago)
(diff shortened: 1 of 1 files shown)


kenoir (approved):

scripts/ss_delete_bag.py
```diff
@@ -0,0 +1,459 @@
...
+    api_name, space, external_identifier, version, date_created
+):
+    """
+    Show the user some information about the bag, and check this is really
+    the bag they want to delete.
+
+    It presents a prompt of the following form:
+
+          This is the bag you are about to delete:
+          Space:        testing
```
kenoir, line 149: Nit: I think this will also display the environment.

alexwlchan: Indeed it will; I realised I had to add that later.

PR #792, opened 2020-11-23
Implement reindexing consistency and retry

kenoir:
These changes were proposed to fix an issue indexing bags that is [likely to be something else](https://github.com/wellcomecollection/storage-service/pull/791). 

They're still a good idea though. 

This change:

- Makes the bag tracker use strongly consistent reads so we can be sure that we never report a bag in an incorrect state
- Makes the bag indexer retry on a non-zero delay if something goes wrong (a zero length delay is unhelpful if a dependent service is temporarily unavailable).

### DIFF
--- a/bag_tracker/src/main/scala/uk/ac/wellcome/platform/archive/bag_tracker/storage/dynamo/DynamoStorageManifestDao.scala
+++ b/bag_tracker/src/main/scala/uk/ac/wellcome/platform/archive/bag_tracker/storage/dynamo/DynamoStorageManifestDao.scala
@@ -29,6 +29,9 @@ class DynamoStorageManifestDao(
   s3Client: AmazonS3
 ) extends StorageManifestDao {
 
+  //  By default reads are eventually consistent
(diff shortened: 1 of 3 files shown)


bag_tracker/src/main/scala/uk/ac/wellcome/platform/archive/bag_tracker/storage/dynamo/DynamoStorageManifestDao.scala
```diff
@@ -37,6 +39,9 @@ class DynamoStorageManifestDao(
   s3Client: AmazonS3
 ) extends StorageManifestDao {
 
+  implicit val consistencyMode: ConsistencyMode =
+    StronglyConsistent
```
alexwlchan, line 43: Worth a comment about why we want strong consistency here?

alexwlchan (approved):

alexwlchan:
For https://github.com/wellcomecollection/platform/issues/4873

PR #794, opened 2020-11-24
Bag deletion only looks at stage

kenoir:
The indentation here is playing tricks with looking at both apis.


alexwlchan (approved):
Oops, good catch!

PR #795, opened 2020-11-24
Use newer versions of all our dependencies

alexwlchan:
For wellcomecollection/platform#4894


kenoir (approved):
Sufficiently un`right`.

PR #796, opened 2020-11-24
Grant blanket read permissions to other AWS accounts

alexwlchan:
For https://github.com/wellcomecollection/platform/issues/4895

Rather than specifying individual roles, which we need to manage, just grant blanket read permission to the entire account.  We can assign permissions to individual roles within those accounts, but without having to keep modifying this block.

We've discussed this change in the past with folks at Digirati, and tested it on some Digirati roles, but apparently never got around to doing it properly.

### DIFF
--- a/terraform/critical_prod/locals.tf
+++ b/terraform/critical_prod/locals.tf
@@ -1,19 +1,10 @@
 locals {
   namespace = "storage"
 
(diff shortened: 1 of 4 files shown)


jamesgorrie (approved):
Very neat.

PR #803, opened 2020-12-08
Refactor a few of the tests in the bag unpacker

alexwlchan:
Some cleanups done while working on https://github.com/wellcomecollection/platform/issues/4911

PR #804, opened 2020-12-08
Add a couple of extra error cases in the bag unpacker

alexwlchan:
Follows #803. For https://github.com/wellcomecollection/platform/issues/4911

Previously, the unpacker has only detected errors at the very start of a file – for example, if you pass it something which isn't a valid tar.gz. There's been no handling of errors that occur *midway* through the file (for example, if a valid tar.gz has been truncated). The user would get a message "Unpacking failed", but no information about how to fix it.

**We now handle several types of failure that occur midway through the file, and provide debugging information in the ingests API for users to see.**

*   **Parsing errors caused by truncated archives.** If the archive is truncated, you can get an IOException midway through unpacking with the message `Error detected parsing the header`. If we get this error, we now report an error to the user:

    > Error trying to unpack the archive at $location - is it the correct format?

    This is the same error message we show if a file is completely unarchive-like.

*  **Duplicate entries within an archive.** While truncating archive files, I discovered that if you truncate an archive in a particular way, and unpack it in S3, the unpacker falls into an infinite loop. Bad!

    That made me realise that the unpacker can potentially lose data if an archive contains duplicate files. It's perfectly legal for a tar archive to contain two different files under the same name. Previously the storage service would pick which ever one comes lost in the archive, but this was by accident rather than by design.

    There's nothing sensible to do in this case (which should we pick?), so if we spot this we now fail the ingest and report an error to the user:

    > The archive at $location is malformed or has a duplicate entry ($entryName)

    (I don't understand exactly why the S3 bug occurs, and I haven't looked in detail. I assume the truncation happened to land on an interesting byte boundary, but it's not that important.)

* **EOF errors caused by truncated archives.** If the archive is truncated, it's possible for the S3 SDK to throw an EOF exception if we've started to write the unpacked file into S3, because the S3 unpacker doesn't read the whole archive at once. If we get this error, we now report an error to the user:

    > Unexpected EOF while unpacking the archive at $Location - is it the correct format?

    This is the same error message we show if we detect an EOF at the start of an archive.


bag_unpacker/src/test/scala/uk/ac/wellcome/platform/archive/bagunpacker/services/s3/S3UnpackerTest.scala
```diff
@@ -185,5 +188,120 @@ class S3UnpackerTest
...
+      *
+      * The EOF exception caught in UnpackerTestCases is caught when we *read* the
+      * compressed bag.  We've read the whole thing into memory, and found it wanting.
+      *
+      * You can also get an EOF exception when you *write* a file from the bag.
+      * In particular, if the S3 unpacker reads a good chunk and starts uploading a file,
+      * then the EOF will be thrown by the uploader when it doesn't get enough bytes from
+      * the next, bad chunk.
+      *
+      * To reproduce this, we have to ensure the final file falls in two different chunks.
```
kenoir, line 257: So many fun ways to fail.

bag_unpacker/src/test/scala/uk/ac/wellcome/platform/archive/bagunpacker/storage/UnarchiverTest.scala
```diff
@@ -98,4 +98,51 @@ class UnarchiverTest extends AnyFunSpec with Matchers with EitherValues {
...
+
+  /** The file for this test was created with the following bash script:
+    *
+    *     dd if=/dev/urandom bs=1024 count=1 > 1.bin
+    *     dd if=/dev/urandom bs=1024 count=1 > 2.bin
+    *     tar -cvf repetitive_non_consecutive.tar 1.bin 2.bin 1.bin
+    *     gzip repetitive_non_consecutive.tar
+    *
+    */
+  it("fails if the archive has non-consecutive repeated entries") {
```
kenoir, line 134: Might be useful to take the description from the PR about why this is a problem and add it in the comment here.

kenoir (approved):

PR #806, opened 2021-01-04
Don't let failures age out in the daily reporter

alexwlchan:
Closes https://github.com/wellcomecollection/platform/issues/4943

This patch modifies the daily Slack reporter to look for any unexplained failures in the reporting index. This means that if an ingest fails, and we don't address it within two days, it will continue to be highlighted in the report. Previously, the ingest would age out and the report would go green again.

When I first read this, it highlighted ~100 ingests with an unexplained failure. I've had a quick look through them, and there's nothing concerning. They look like bugs that we've since fixed or improved the error reporting on. I've hard-coded a list of these failures to draw a line under them; new unexplained failures we can fix or ignore as they arise.

Separately, I've modified the reporter so it only runs on weekdays. Nobody should be looking at the report on the weekend, so we don't need to run it. When the report runs on Monday morning, it will pick up any ingests that got updated over the weekend.

I've deployed this change.


monitoring/daily_reporter/src/ingests.py
```diff
@@ -72,3 +99,116 @@ def get_dev_status(ingest):
             return "stalled"
         else:
             return "processing"
+
+
+def is_known_failure(ingest_id):
+    return ingest_id in {
+        # digitised/b12948421, failed with an unknown unpacker error
+        # Reingested as f9850ea4-28b3-4ef2-a385-d0227fb5167a
+        "dda879ab-ba7b-4e87-a34b-6511e6e1457b",
```
kenoir, line 108: I'm not sure I like this ... it's an infinitely growing list, that relies on a dev properly commenting things to make it sensible.

kenoir:
I'd prefer to keep the current functionality but still have a reference for old failures - being able to highlight recent failures is useful. 

I have also just spent a few weeks where the weekend reporting is useful.

monitoring/daily_reporter/src/elasticsearch.py
```diff
@@ -129,4 +131,48 @@ def get_interesting_ingests(es_client, *, index_name, days_to_fetch):
...
+            "failed (user error)",
+            "failed (known error)",
+        }:
+            continue
+        else:
+            ingests[hit["_id"]] = failed_ingest
+
+        # If there are more than 10000 results, that suggests there are more
+        # that we didn't get.  Flag this to the caller.
+        if resp.json()["hits"]["total"]["value"] >= 10000:
```
jamesgorrie, line 175: You could use [scroll](https://elasticsearch-py.readthedocs.io/en/7.10.0/api.html?highlight=scroll#elasticsearch.Elasticsearch.scroll) if you wanted more

alexwlchan: \*nod*

That means adding functionality that I'd have to test and maintain, and right now it's not an issue – we never actually hit the 10000 limit, but I don't want it to fail silently if we do.

alexwlchan:
> I'd prefer to keep the current functionality but still have a reference for old failures - being able to highlight recent failures is useful.
>
> I have also just spent a few weeks where the weekend reporting is useful.

I'll have another think about this.

alexwlchan:
I’ve rolled this back until I can think about it some more.

PR #807, opened 2021-01-04
Verify that our bags include a bag declaration

alexwlchan:
The first "required element" [in the BagIt spec](https://tools.ietf.org/html/rfc8493#section-2.1.1) is a "Bag Declaration", aka bagit.txt:

> The "bagit.txt" tag file MUST consist of exactly two lines in this order:
>
> ```
> BagIt-Version: M.N
> Tag-File-Character-Encoding: ENCODING
> ```

We didn't have any code that checked for the presence of this file; now we do. I'm running a script to check we have an appropriate bagit.txt file in all the existing bags.


alexwlchan:
I've had a look through the existing bags, and I'm happy they seem to have the correct declaration.

kenoir (approved):

PR #808, opened 2021-01-05
Verify that payload files are in data/, tag files are in the root

alexwlchan:
Second part of https://github.com/wellcomecollection/platform/issues/4931, follows #807 

I think all our bags already follow these two rules. The "data/" rule is required by the BagIt spec and one we've not been rigidly enforcing; tag files in the root is a convention we've adopted that we should discuss if we ever want to break (rather than letting it slip through quietly).

### DIFF
--- a/bag_verifier/src/test/scala/uk/ac/wellcome/platform/archive/bagverifier/verify/steps/VerifyFilenamesTest.scala
+++ b/bag_verifier/src/test/scala/uk/ac/wellcome/platform/archive/bagverifier/verify/steps/VerifyFilenamesTest.scala
@@ -10,10 +10,7 @@ import uk.ac.wellcome.platform.archive.common.bagit.models.{
 }
 import uk.ac.wellcome.platform.archive.common.verify.{ChecksumValue, MD5}
 
-class VerifyFilenamesTest
-    extends AnyFunSpec
-    with Matchers
-    with EitherValues {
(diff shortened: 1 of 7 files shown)


bag_verifier/src/test/scala/uk/ac/wellcome/platform/archive/bagverifier/verify/steps/VerifyFilenamesTest.scala
```diff
@@ -0,0 +1,114 @@
...
+      )
+    }
+  }
+
+  describe("verifyPayloadFilenames") {
+    it("allows filenames that start with data/") {
+      val manifest = PayloadManifest(
+        checksumAlgorithm = MD5,
+        entries = Map(
+          BagPath("data/README.txt") -> ChecksumValue("123"),
```
kenoir, line 47: Do we really need this check?

kenoir (approved):

bag_verifier/src/test/scala/uk/ac/wellcome/platform/archive/bagverifier/verify/steps/VerifyFilenamesTest.scala
alexwlchan: The BagIt spec requires this, so I think so.

It also enforces an assumption that's only been implicit, which is that "BagIt metadata" and "interesting files" live in two separate directories.

PR #824, opened 2021-03-22
Update scala-libs to v26.4.3

alexwlchan:
This gets us the slightly more efficient LockingService and SQSStream, plus we get to ditch a pile of Elasticsearch code that's now available in the common libs.


alicefuzier (approved):
Nice!

PR #825, opened 2021-03-31
Use the LargeStreamReader to increase reliability a little further

alexwlchan:
Last night, we had a failure in one of the S3 bag verifiers because the connection timed out midway through reading an object. The object in question was multiple gigabytes in size (an MXF file).

This patch is my attempt to make this sort of failure less likely in future.

---

A quick refresher:

A timeout when reading an object isn't an uncommon failure mode: the larger the file, the longer you need to hold the screen open, the higher the risk of something dropping in the middle. To get around this in the bag _unpacker_, we divide large objects into smaller pieces, and then stitch the pieces back together in memory. Additionally, we read all the bytes for each piece immediately, so we're not holding a connection open for a long time. We also have retry logic around reading each piece.

I have a blog post about this: https://alexwlchan.net/2019/09/streaming-large-s3-objects/ and you can see our current implementation here: https://github.com/wellcomecollection/scala-libs/blob/master/storage/src/main/scala/uk/ac/wellcome/storage/services/LargeStreamReader.scala

We already use this "break the object into smaller pieces" reader for the Azure verifier, but we never went back and added it to S3. This reader should be a direct drop-in for the "read it all in one go" that we were previously using, but hopefully a bit more reliable.

---

The "more correct" fix would be to modify the verifier so that it spots a retryable error midway through reading an object, and doesn't fail the entire ingest. I've added a comment in the rough code location where we'd add this, but I'm not going to do it today. This code already works pretty well, and adding even more retry logic would be a lot of work for very marginal gain.

### DIFF
--- a/bag_verifier/src/main/scala/uk/ac/wellcome/platform/archive/bagverifier/fixity/azure/AzureFixityChecker.scala
+++ b/bag_verifier/src/main/scala/uk/ac/wellcome/platform/archive/bagverifier/fixity/azure/AzureFixityChecker.scala
@@ -48,6 +48,10 @@ object AzureFixityChecker {
     //      java.util.concurrent.TimeoutException: Did not observe any item or terminal
     //      signal within 60000ms in 'map' (and no fallback has been configured))
     //
+    // Note: in hindsight, this may have been less to do with Azure, and more to do with
+    // the constrained bandwidth on the NAT instance we were using to talk to Azure when
+    // we did the initial migration.  The managed NAT Gateway has more capacity.
+    //
     // I had the idea to shrink the size of the buffer by looking at AzCopy, a utility
     // for downloading large blobs from Azure… as files.  The actual download code seems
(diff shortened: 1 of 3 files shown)


bag_verifier/src/main/scala/uk/ac/wellcome/platform/archive/bagverifier/fixity/azure/AzureFixityChecker.scala
```diff
@@ -57,7 +62,9 @@ object AzureFixityChecker {
     // not by much.
     //
     // This bufferSize has successfully verified a blob which was 166 GiB in size.
-    val streamReader = AzureLargeStreamReader(bufferSize = 16 * 1024 * 1024) // 16 MB
+    val streamReader = AzureLargeStreamReader(
+      bufferSize = 16 * FileUtils.ONE_MB
```
kenoir, line 66: ❤️

kenoir (approved):

PR #835, opened 2021-05-06
Use latest http libs

kenoir:
Use contextUrl and withApp in fixtures.


kenoir:
For https://github.com/wellcomecollection/platform/issues/5162

PR #838, opened 2021-05-19
Move cognito user pool from platform-infra

kenoir:
The cognito user pool is used exclusively by the storage service so we'll move it here to keep the code together.

All necessary state has been imported. 

Removal of state from platform-infra is in: https://github.com/wellcomecollection/platform-infrastructure/pull/179


alexwlchan (changes requested):
You also need to export the Cognito user pool ARN and update the storage-service stacks to read the value from this configuration: https://github.com/search?type=Code&q=org:wellcomecollection+cognito_user_pool_arn

PR #864, opened 2021-07-01
Make-ing the build system be less magic

alexwlchan:
I want the storage-service to be publishing images to ECR Public as well as ECR Private, for https://github.com/wellcomecollection/platform/issues/5226.

This means fiddling with the build systems, which means fiddling with the Makefiles, which is a world of complexity and pain. I have thus decided to rip out all the Make and replace it with a handful of top-level scripts in the "builds" folder.

Benefits/reasoning:

* We have plenty of shell/Python elsewhere in the build system, so we're all sharp in those tools. Make is mostly indirection and icing, and not something we look at very often. Expunging it gets a build system we'll all be more confident using and changing.
* We already have a source of truth for Scala apps in the repo: sbt itself. The sbt information is serialised into neat JSON files in `.sbt_metadata`, so let's use that instead of having humans maintain the information in a collection of Makefiles.
* Also, computers can programatically determine the existence of a docker-compose/Dockerfile. Let's do that, instead of having humans manage the list.
* If you can work out what the Makefiles are doing, you'll see they run a collection of Docker images that haven't been updated in years – and finding out what those Docker images are doing is quite hard. I've replaced them with inline shell scripts that involve less magic and are easier to follow.

   I've copied the "publish Lambda ZIP" script from the Docker image, and the other scripts are based on what those Docker images were doing (when you eventually found them).

Assuming people think this is a good idea, I'll roll this pattern out to our other repos.


.buildkite/scripts/commands.py
```diff
@@ -27,9 +27,9 @@ def _subprocess_run(cmd, exit_on_error=True):
     return "\n".join(output).strip()
 
 
-def make(*args):
+def run_build_script(name, *args):
     """Run a Make command, and check it completes successfully."""
```
kenoir, line 31: Not any more.

kenoir (approved):

PR #867, opened 2021-07-05
Tidy up a bunch of rough edges for running the storage-service in a new account

alexwlchan:
For https://github.com/wellcomecollection/platform/issues/5206

* Make sure we actually define the bag register repo in Terraform
* There's a new "working storage" module inside the `stack` module, which no longer hard-codes the "wellcomecollection" prefix
* You can specify an arbitrary container registry and container tag – e.g. ECR Public and the "latest" tag, or ECR Private and the "env.prod" tag


terraform/infra/ecr.tf
```diff
@@ -2,6 +2,7 @@ locals {
   repository_names = [
     "bags_api",
     "bag_indexer",
+    "bag_register",
```
kenoir, line 5: errr, how did we miss this?

kenoir (approved):

terraform/infra/ecr.tf
alexwlchan: I think what happened is that we rejigged the ECR repo definitions at some point, and when we were rearranging the Terraform state the bag register got missed.

PR #871, opened 2021-07-06
Add a document explaining how we route requests from an HTTP request to app containers

alexwlchan:
Rendered version: https://github.com/wellcomecollection/storage-service/blob/request-routing/docs/how-are-requests-routed.md

Part of https://github.com/wellcomecollection/platform/issues/5206

It's possible to work this out by staring at the Terraform, but API Gateway in particular makes this quite non-obvious, and having it written out in a one-pager feels like something useful to have.


kenoir:
Nice. It might help to include some references to how ports are used at each step, as that will be relevant for reconciling the doc with the code. For example at nginx / vpc link stages we are routing from one port to another - you might want to highlight that.

alexwlchan:
> It might help to include some references to how ports are used at each step, as that will be relevant for reconciling the doc with the code. For example at nginx / vpc link stages we are routing from one port to another - you might want to highlight that.

Good suggestion, done.

alexwlchan:
I forgot to mention Route 53/DNS entries. >.<

alexwlchan:
Superseded by another PR.

PR #873, opened 2021-07-06
Add a "key technologies" document

alexwlchan:
Rendered version: https://github.com/wellcomecollection/storage-service/blob/technologies-doc/docs/technologies.md

This is a first pass at a reading list of topics for anybody who wants to work on the storage service – a one-page guide to the terms we throw around regularly, meant for a software developer who hasn't encountered these specific technologies before. Anything else I should add?


docs/technologies.md
```diff
@@ -0,0 +1,80 @@
...
+    </td>
+  </tr>
+  <tr>
+    <td>
+      <img src="./images/Arch_AWS-Fargate_64@5x.png">
+    </td>
+    <td>
+      <strong>ECS/Fargate</strong> – a serverless container runtime.
+      We package our services in Docker images, and then Fargate actually runs the containers, without us having to provision VMs/servers to run them on.
+      Fargate is a subservice of Elastic Container Service, or ECS.
```
jamesgorrie, line 46: Worth mentioning cost savings?

docs/technologies.md
```diff
@@ -0,0 +1,80 @@
...
+      <img src="./images/Arch_Amazon-Simple-Queue-Service_64@5x.png">
+      <img src="./images/Arch_Amazon-Simple-Notification-Service_64@5x.png">
+    </td>
+    <td>
+      <strong>SNS/SQS</strong> – inter-app message queues.
+      Our apps form a pipeline: an app gets a message from an SQS queue, does some work, then sends an SNS notification to the next app.<br/><br/>
+      We use SNS <em>topics</em> and SQS <em>queues</em>.
+      Queues can subscribe to the output of topics, in a many-to-many relationship.
+      This allows one app to send messages to multiple apps, or one app to receive messages from multiple apps.
+    </td>
```
jamesgorrie, line 37: THe structure of this bit seems a bit incongruent with the rest, which seems to follow "Thing - nice simple description of thing. Why we use said thing".

A small image of a pipeline could speak a million words too - I'm sure I've seen one.

alexwlchan: I'm thinking I might do a whole other bit about messaging tbh, to explain the one-to-many and many-to-one stuff, and also autoscaling. You're right, it's not a good fit here – a "read more" link would be better.

docs/technologies.md
alexwlchan: What were you thinking? The Fargate cost story is sufficiently complicated that I'm not sure what I could write in a sentence or two.

docs/technologies.md
alexwlchan: How about this? https://github.com/wellcomecollection/storage-service/blob/technologies-doc/docs/inter-app-messaging.md

alexwlchan:
Superseded by another PR.

PR #875, opened 2021-07-07
Run the storage-service in docker-compose

alexwlchan:
For a slightly easier local demo.

This seems within the realm of possibility now. So far:

- ✅ Starting the localstack container to simulate AWS services, run Terraform to define infrastructure
- ✅ Start the ingests tracker and have it write ingests to DynamoDB
- ✅ Start the ingests worker and have it read from the SQS queue

Current issues:

- ⚠️ Creating SNS topics doesn't work if I blow away the localstack container:

    ```
    Error: error reading SNS Topic (arn:aws:sns:eu-west-1:000000000000:callback_notifications): :
    aws_1              | 	status code: 404, request id:
    ```

- ⚠️ I need to the ingests tracker to publish on 0.0.0.0 so it can be reached by the ingests worker.

### DIFF
--- a/demo/docker/.gitignore
+++ b/demo/docker/.gitignore
@@ -1,0 +1,1 @@
+localstack
--- a/demo/docker/docker-compose.yml
+++ b/demo/docker/docker-compose.yml
@@ -19,7 +19,7 @@ services:
       - "DATA_DIR=/tmp/localstack/data"
 
   ingests_tracker:
-    image: "ingests_tracker:43c9ad494468929d4a7edb98da7b93f72ac144af"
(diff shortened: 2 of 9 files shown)


alexwlchan:
A thought on this: you can precreate S3 buckets and the like in localstack: https://stackoverflow.com/q/53619901/1558022

alexwlchan:
This PR was originally created to help external orgs who wanted to use the storage service run it in their environment.

It never actually went anywhere, nobody except Wellcome is using it, and I’m not sure Wellcome is currently in a position to support external orgs even if they did want to use it. Closing.

PR #879, opened 2021-07-09
Another round of documentation updates

alexwlchan:
Mostly what this adds is a skeleton I can use to build out the rest of the documentation as I write it.

Rendered version: https://github.com/wellcomecollection/storage-service/blob/diataxis2/README.md


docs/README.md
```diff
@@ -11,149 +11,59 @@ Each bag is identified with a **space** (a broad category) an **external identif
...
 
-## Releasing
+These topics explain how the storage service work, and why it's designed in the way it is:
 
-```sh
-# Ensure you have the correct AWS credentials in scope
-# before running!
+-   [Detailed architecture: what do the different services do?]
+-   [How identifiers work in the storage service](explanations/identifiers.md)
+-   [How files are laid out in the underlying storage](explanations/file-layout.md)
```
kenoir, line 54: I don't think `explanations/file-layout.md` is in this PR.

docs/developers/inter-app-messaging.md
```diff
@@ -0,0 +1,56 @@
+# Inter-app messaging with SQS and SNS
```
kenoir, line 1: Do you really think we need to explain this? Could we not point to resources about the pubsub pattern & the AWS docs?

It's a good explanation, I suppose whether or not we need it depends on the audience.

kenoir: Whether or not we do, perhaps worth referencing some [nice explanation of pubsub](https://en.wikipedia.org/wiki/Publish%E2%80%93subscribe_pattern).

kenoir (approved):

docs/README.md
alexwlchan: Oops, probably because I haven't written it yet. That's next.

docs/developers/inter-app-messaging.md
alexwlchan: I'm assuming an audience who isn't super familiar with the internals of AWS – this is (like a bunch of the other documentation) meant to be a bit of a reading list.

I do think the SQS-backed autoscaling is somewhat unusual and worth mentioning.

PR #888, opened 2021-07-15
Continue to make the storage-service more region agnostic

alexwlchan:
For https://github.com/wellcomecollection/platform/issues/5241

This patch:

* Will delete secrets immediately when the demo stack is torn down
* Use Scala apps that don't hard-code their region as `eu-west-1`
* Use a version of the SQS queue module that doesn't need a region variable set (https://github.com/wellcomecollection/terraform-aws-sqs/pull/7)

### DIFF
--- a/demo/terraform/demo_stack/stack.tf
+++ b/demo/terraform/demo_stack/stack.tf
@@ -96,6 +96,6 @@ module "stack" {
 
   app_containers = {
     container_registry = "public.ecr.aws/y1p3h6z3"
-    container_tag      = "latest"
+    container_tag      = "ref.c3b7e5a1557ba0115a31ad06e6a7edd5ad143a64"
   }
 }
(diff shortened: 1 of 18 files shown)


demo/terraform/demo_stack/stack.tf
```diff
@@ -1,10 +1,10 @@
 module "stack" {
-  source = "github.com/wellcomecollection/storage-service.git//terraform/modules/stack?ref=caab5b2"
+  source = "github.com/wellcomecollection/storage-service.git//terraform/modules/stack?ref=52e10ed"
```
kenoir, line 2: Can we create a tag specifically for the demo?

demo/terraform/demo_stack/stack.tf
```diff
@@ -96,6 +96,6 @@ module "stack" {
 
   app_containers = {
     container_registry = "public.ecr.aws/y1p3h6z3"
-    container_tag      = "latest"
+    container_tag      = "ref.c3b7e5a1557ba0115a31ad06e6a7edd5ad143a64"
```
kenoir, line 99: Similar to above can we tag a particular container image to denote that it is the demo image?

kenoir: It's not really "demo" in this case 🤔

kenoir (approved):

demo/terraform/demo_stack/stack.tf
alexwlchan: Good suggestion. I haven't yet because the demo is still in flux, but one to revisit later.

PR #892, opened 2021-07-16
Create an autogenerated guide to projects in the repository

alexwlchan:
The idea of this patch is to make it easier for new developers to find their way round the repo, but without writing a long guide that will fall out-of-date. Here's how it works:

* We write a one-line description of the service in `build.sbt` when we define the project
* This description gets copied into the JSON serialisation of the sbt metadata that we use for short-circuiting in CI
* A Python script picks up this description as part of the `autoformat` task and renders a list of projects

See rendered version: https://github.com/wellcomecollection/storage-service/blob/self-documenting-sbt/docs/developers/project-guide.md


kenoir (approved):
That is neat.

kenoir:
I think it would be slightly unexpected to a Scala dev coming across this repo that we use Python to produce this information, but this isn't the first example of that in our projects, and it works.

alexwlchan:
> I think it would be slightly unexpected to a Scala dev coming across this repo that we use Python to produce this information, but this isn't the first example of that in our projects, and it works.

It could be written in Scala, yeah. My initial thought was something much more complicated, where they were ordered by dependency – i.e. core dependencies first, apps later – and we already have the Python code to deal with that for the CI stuff.

I didn't implement that because it was complicated, but it might be useful in the pipeline repo, where we have a much less flat structure. You can only really do that when you consider the project as a whole.

Issue #894, opened 2021-07-16
Support bags that use Windows line endings


alexwlchan:
If you upload a Bag Declaration (`bagit.txt`) like this:

```
BagIt-Version: 0.97\r\nTag-File-Character-Encoding: UTF-8\r\n
```

you get a mostly unhelpful message from the storage service:

> Verification (pre-replicating to archive storage) failed - Error loading Bag Declaration (bagit.txt): not correctly formatted

If the checksums are correct, there's no reason we can't support Windows-style line endings (and indeed, we do in most other parts of the storage service).

PR #895, opened 2021-07-16
Support Windows line endings in the bag verifier

alexwlchan:
For https://github.com/wellcomecollection/storage-service/issues/894

Previously we had an overly strict regex for matching the Bag Declaration (bagit.txt):

```scala
private val declaration = new Regex(
  "BagIt-Version: \\d\\.\\d+\nTag-File-Character-Encoding: UTF-8\n?"
)
```

Now we defer to Scala to split lines in bagit.txt for us (`String.lines`), so we don't have to worry about supporting different line endings.  This means we get support for Windows line endings for "free".

This approach also means we can provide slightly better error messages when we don't accept the Bag Declaration as supplied, e.g. "encoding must be UTF-8" or "not correctly formatted".

Once this patch is merged, I'll try a Windows bag in the staging environment and see if anything else breaks; if not, I'll close the ticket.

### DIFF
--- a/bag_verifier/src/main/scala/weco/storage_service/bag_verifier/verify/steps/VerifyBagDeclaration.scala
+++ b/bag_verifier/src/main/scala/weco/storage_service/bag_verifier/verify/steps/VerifyBagDeclaration.scala
@@ -6,6 +6,7 @@ import weco.storage.streaming.InputStreamWithLength
 import weco.storage.{Identified, Location, NotFoundError, Prefix}
 import weco.storage.streaming.Codec._
 
+import scala.io.Source
 import scala.util.matching.Regex
 
 trait VerifyBagDeclaration[BagLocation <: Location, BagPrefix <: Prefix[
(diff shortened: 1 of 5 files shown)


kenoir (approved):

alexwlchan:
Quick guide to the files in this PR for anybody who's not worked on the storage service before:

- BagBuilder creates example bags that we can use for testing. It also lets us override a bunch of behaviours, e.g. creating bags that have bad checksums.

- Each of the classes in [`bag_verifier/verify/steps`](https://github.com/wellcomecollection/storage-service/tree/cfb6ac2d734ac5dc1d234d4cf317d746fc4b6968/bag_verifier/src/main/scala/weco/storage_service/bag_verifier/verify/steps) checks a particular part of the bag, and we have detailed tests for that particular check

- The BagVerifier is a trait that comes in multiple implementations (a standalone bag in S3, a replicated bag in S3, a replicated bag in Azure), and BagVerifierTestCases contains a set of generic test cases that should apply to every implementation. We then have one implementation of the TestCases for each implementation of the core trait, so every test case is run against every implementation.

   This is a pretty common pattern in our codebase: a base trait, a TestCases trait, and one test implementation for every implementation of the trait.

Issue #900, opened 2021-07-19
The storage service should support using checksums that aren't SHA-256


alexwlchan:
Quoting [the BagIt RFC](https://datatracker.ietf.org/doc/html/rfc8493#section-2.4):

> Starting with BagIt 1.0, bag creation and validation tools MUST
   support the SHA-256 and SHA-512 algorithms [RFC6234] and SHOULD
   enable SHA-512 by default when creating new bags.  For backwards
   compatibility, implementers SHOULD support MD5 [RFC1321] and SHA-1
   [RFC3174].  Implementers are encouraged to simplify the process of
   adding additional manifests using new algorithms to streamline the
   process of in-place upgrades.

Right now we only support SHA-256. We should add support for other checksum algorithms.

alexwlchan:
Failure modes we need to consider:

* There's no SHA-256/SHA-512 manifest
* The manifests have different entries
* One manifest is valid but the other isn't
* We have inconsistent manifests, e.g. a MD5 manifest and a SHA256 tag manifest

PR #901, opened 2021-07-19
Support using checksums that aren't SHA-256

alexwlchan:
Closes https://github.com/wellcomecollection/storage-service/issues/900

> Starting with BagIt 1.0, bag creation and validation tools MUST
support the SHA-256 and SHA-512 algorithms [RFC6234] and SHOULD
enable SHA-512 by default when creating new bags. For backwards
compatibility, implementers SHOULD support MD5 [RFC1321] and SHA-1
[RFC3174]. Implementers are encouraged to simplify the process of
adding additional manifests using new algorithms to streamline the
process of in-place upgrades.

This is taking us towards the BagIt 1.0 definition of a "complete" bag:

>    A _complete_ bag MUST meet the following requirements:
>
>   1.  Every required element MUST be present (see Section 2.1).
>   2.  Every file listed in every tag manifest MUST be present.
>   3.  Every file listed in every payload manifest MUST be present.
>   4.  For BagIt 1.0, every payload file MUST be listed in every payload
       manifest.  Note that older versions of BagIt allowed payload
       files to be listed in just one of the manifests.
>   5.  Every element present MUST conform to BagIt 1.0.
>
>   A _valid_ bag MUST meet the following requirements:
>
>   1.  The bag MUST be _complete_.
>   2.  Every checksum in every payload manifest and tag manifest has
>       been successfully verified against the contents of the
>       corresponding file.

Currently we only support SHA-256 manifests: we'll ignore checksums in other manifest files, and if a bag doesn't include SHA-256 checksums we're unable to verify/register it.

This patch implements full support for SHA-256 and SHA-512, and partial support for MD5 an SHA-1. In particular:

* When a bag is verified, we'll look at the checksums in every manifest. Previously we would ignore them.
* The bag register will choose the "best" checksum to use in the storage manifest – SHA-512 if available, otherwise SHA-256.
* We don't support bags that only have MD5 or SHA-1 checksums – these are for backwards-compatibility only. You need to supply at least one of SHA-256 or SHA-512.
* It's not possible to store bags that don't have SHA-256 checksums, if the caller provides SHA-512 instead.

In a bit more detail, we add the following new checks to follow the BagIt spec:

* If the checksum in any manifest is wrong, we fail the bag.
* If any manifest is missing files, we fail the bag.
* If the same checksums aren't used consistently, we fail the bag. e.g. a bag with `manifest-md5.txt` and `manifest-sha256.txt` would fail.

The bag verifier is an incredibly important service – a mistake in this code would be catastrophic. The patch is too big to safely review as-is, so I'm opening this PR to get high-level feedback. Is this design sensible? Have I missed anything? Is there anything you'd like to double-check?

I'll gradually break this up into smaller PRs for proper review.

### DIFF
--- a/bag_register/src/main/scala/weco/storage_service/bag_register/services/S3StorageManifestService.scala
+++ b/bag_register/src/main/scala/weco/storage_service/bag_register/services/S3StorageManifestService.scala
@@ -114,18 +114,26 @@ class S3StorageManifestService(implicit s3Client: AmazonS3) extends Logging {
     *     by the BagReader. class
     *
     */
-  private def chooseAlgorithm(tagManifest: NewTagManifest): Try[HashingAlgorithm] = {
+  private def chooseAlgorithm(
+    tagManifest: NewTagManifest
+  ): Try[HashingAlgorithm] = {
     val (_, checksum) = tagManifest.entries.head
 
     val algorithms =
       Seq(SHA512, SHA256)
-        .map { ha => (ha, checksum.getValue(ha)) }
+        .map { ha =>
+          (ha, checksum.getValue(ha))
+        }
         .collect { case (ha, Some(_)) => ha }
 
     algorithms.headOption match {
       case Some(algo) => Success(algo)
       case None =>
-        Failure(new Throwable(s"Unable to choose algorithm from tag manifest: $tagManifest"))
+        Failure(
+          new Throwable(
+            s"Unable to choose algorithm from tag manifest: $tagManifest"
+          )
+        )
     }
   }
 
--- a/bag_register/src/main/scala/weco/storage_service/bag_register/services/S3StorageManifestService.scala
+++ b/bag_register/src/main/scala/weco/storage_service/bag_register/services/S3StorageManifestService.scala
@@ -2,7 +2,6 @@ package weco.storage_service.bag_register.services
 
 import java.net.URI
 import java.time.Instant
-
 import com.amazonaws.services.s3.AmazonS3
 import grizzled.slf4j.Logging
 import weco.storage_service.bagit.models._
@@ -17,6 +16,7 @@ import weco.storage.services.s3.S3SizeFinder
 import weco.storage.store.Readable
 import weco.storage.store.s3.S3StreamStore
 import weco.storage.streaming.InputStreamWithLength
+import weco.storage_service.verify.{HashingAlgorithm, SHA256, SHA512}
 
 import scala.util.{Failure, Success, Try}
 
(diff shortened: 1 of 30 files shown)


bag_register/src/main/scala/weco/storage_service/bag_register/services/S3StorageManifestService.scala
```diff
@@ -224,8 +266,13 @@ class S3StorageManifestService(implicit s3Client: AmazonS3) extends Logging {
             }
         }
 
+        // This lookup should never fail -- we should already have validated in
+        // the BagReader class and the bag verifier that if an algorithm is used,
+        // it is used to describe every file in the bag.
+        val checksum = multiChecksum.getValue(algorithm).get
```
kenoir, line 272: Perhaps we should still throw a sensible error though?

bag_register/src/main/scala/weco/storage_service/bag_register/services/S3StorageManifestService.scala
```diff
@@ -96,6 +100,43 @@ class S3StorageManifestService(implicit s3Client: AmazonS3) extends Logging {
...
     } yield storageManifest
 
+  /** Currently the StorageManifest model only supports a single algorithm, but
+    * a bag can contain multiple manifest files for different algorithms.
+    *
+    * Choose the algorithm to use for this storage manifest.
+    *
+    * Implementation notes:
+    *
+    *   - We know the tag manifest is empty because it has to describe
```
kenoir, line 110:
```suggestion
    *   - We know the tag manifest is not empty because it has to describe
```

common/src/main/scala/weco/storage_service/bagit/models/MultiChecksumValue.scala
```diff
@@ -0,0 +1,43 @@
...
+  md5: Option[T] = None,
+  sha1: Option[T] = None,
+  sha256: Option[T] = None,
+  sha512: Option[T] = None
+) {
+  if (md5.isEmpty && sha1.isEmpty && sha256.isEmpty && sha512.isEmpty) {
+    throw MultiChecksumException.NoChecksum
+  }
+
+  // We support MD5 and SHA1 for backwards compatibility (see RFC 8493 § 2.4)
```
kenoir, line 22: I wonder if linking directly to the IETF version of the manifest would be useful (here and elsewhere).

kenoir:
I wonder if we should check that the Bagit spec version is >= 1.0 as part of this as we are implicitly requiring that here.

kenoir:
This design seems sensible to me. 

Interesting to compare this implementation to https://github.com/LibraryOfCongress/bagit-java 

I wonder if we should be pushing our verification code in a direction that makes it less dependent underlying storage provider (in our case we require S3, the LOC version requires a filesystem).

jtweed:
Sorry a bit late to this, but it seems like a really good idea to me. This bit of work is turning into a good opportunity for us to make sure we are compliant with the 1.0 spec. Can't speak to the code, but the logic seems sensible.

PR #902, opened 2021-07-20
Add the models to support checksums from multiple manifests

alexwlchan:
The first piece extracted from https://github.com/wellcomecollection/storage-service/pull/901

It introduces the `MultiManifestChecksum` model for recording the checksums from different payload manifests, and does some renaming and rename spacing to use the word "checksum", which is the term more commonly used in the BagIt RFC. These models aren't used anywhere yet, but they lay the foundation for the next bit of work.

Just focusing on the models has let me tidy them up a bit, so it should be easier to add new checksum algorithms if we ever decide to do so.

### DIFF
--- a/common/src/main/scala/weco/storage_service/checksum/ChecksumAlgorithm.scala
+++ b/common/src/main/scala/weco/storage_service/checksum/ChecksumAlgorithm.scala
@@ -1,4 +1,4 @@
-package weco.storage_service.verify
+package weco.storage_service.checksum
 
 import org.apache.commons.codec.digest.MessageDigestAlgorithms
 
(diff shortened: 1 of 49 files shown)


common/src/main/scala/weco/storage_service/checksum/ChecksumAlgorithm.scala
```diff
@@ -0,0 +1,60 @@
...
+  val value: String = MessageDigestAlgorithms.SHA_256
+}
+
+case object SHA1 extends ChecksumAlgorithm {
+  val value: String = MessageDigestAlgorithms.SHA_1
+
+  override val isForBackwardsCompatibilityOnly: Boolean = true
+}
+
+case object MD5 extends ChecksumAlgorithm {
```
kenoir, line 56: Nice and extensible 👏

common/src/test/scala/weco/storage_service/checksum/ChecksumAlgorithmsTest.scala
```diff
@@ -0,0 +1,22 @@
...
+import org.scalatest.funspec.AnyFunSpec
+import org.scalatest.matchers.should.Matchers
+import weco.storage_service.fixtures.ReflectionHelpers
+
+class ChecksumAlgorithmsTest
+    extends AnyFunSpec
+    with Matchers
+    with ReflectionHelpers {
+
+  // This is meant to help us remember to update the list if/when we add
```
kenoir, line 12: Nice touch.

kenoir (approved):

PR #904, opened 2021-07-21
Ensure API Gateway Deployments contain all resources

h4l:
This PR fixes an intermittent issue that can occur on initial Terraform deployments — the API Gateway Deployment can be created before all of the API resources are defined, so things like authentication or API resources/methods may not be present in the Deployment. It also avoids a theoretical problem in that future changes to API resources currently don't trigger the creation of a new Deployment, so API changes wouldn't be visible without manual intervention to update the Deployment.

The fix is to make the relationship between API Gateway resources (`aws_api_gateway_method`,
`aws_api_gateway_resource`, etc) explicit, so that Terraform creates/updates them prior to `aws_api_gateway_deployment`, and also re-creates the Deployment if the resources it indirectly contains are modified. (Terraform doesn't know about these relationships by default as the sub-resources reference the `api_gateway_rest_api`, not the `aws_api_gateway_deployment`).

I've implemented this via the `triggers` argument of the Deployment, much like the example in the terraform docs: https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/api_gateway_deployment

---

Additionally, the "v1" Stage is now explicitly managed as a `api_gateway_stage` resource, rather than having the `aws_api_gateway_deployment` create one automatically via `stage_name`. This is the approach recommended in the `aws_api_gateway_deployment` docs linked above. 

**However, applying this change to an existing deployment re-trigger the original error seen on first apply**: 

> Error creating API Gateway Stage: ConflictException: Stage already exists

The commit message for the change describes why this occurs. To avoid it, the "v1" Stage needs to be manually deleted before the apply. After that the conflict will no longer occur. (I couldn't find a way to avoid this need for manual intervention, other than by renaming the stage, or renaming the `api_gateway_rest_api`, neither of which seemed desirable.)

This explicit Stage management is not required for the resource ordering/Deployment re-creation, so we could drop this aspect of the PR. The warning box in the `api_gateway_deployment` docs describe why it's recommended to explicitly manage the Stage (it avoids a brief period of downtime when the Deployment is re-created).

### DIFF
--- a/demo/terraform/README.md
+++ b/demo/terraform/README.md
@@ -23,7 +23,7 @@ It spins up real resources in your AWS account and you will be billed for them.
 
     ```hcl
     module "demo_stack" {
-      source = "github.com/wellcomecollection/storage-service.git//demo/terraform/demo_stack?ref=b155543"
+      source = "github.com/wellcomecollection/storage-service.git//demo/terraform/demo_stack?ref=5a61182"
 
       namespace       = "..."  # e.g. weco-dams-prototype
       short_namespace = "..."  # e.g. weco -- this should be 5 chars or less
--- a/demo/terraform/demo_stack/metadata_stores.tf
+++ b/demo/terraform/demo_stack/metadata_stores.tf
@@ -1,5 +1,5 @@
 module "metadata_stores" {
-  source = "github.com/wellcomecollection/storage-service.git//terraform/modules/critical/metadata_stores?ref=52e10ed"
+  source = "github.com/wellcomecollection/storage-service.git//terraform/modules/critical/metadata_stores?ref=2586bca81"
 
   namespace = var.namespace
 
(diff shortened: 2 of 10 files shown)


alexwlchan (approved):
Thanks @h4l, this looks good to me! I'm going to spin up a fresh stack with these changes to check they're all okay, then I'll merge this. ✨ 

(CI hasn't run on this PR because it's not configured to run on PRs from forks. I want to double-check with the other Wellcome devs how we want to handle that, but in the meantime I've tested this locally and it looks okay. 👍)

alexwlchan:
Yup, this created successfully for me. Thanks @h4l.

Congrats on the first contribution to the storage service from somebody not employed by or contracted to Wellcome! 🍰

h4l:
Awesome, cheers @alexwlchan 🍻

PR #910, opened 2021-07-23
Read multiple checksums in the BagReader

alexwlchan:
For #900, follows #909.

This patch means that whenever we read a bag from S3 or Azure, we'll read all the available manifests and check they're consistent. The applications will only look at the SHA256 manifest, but the other manifests will be sitting in-memory waiting to be used.

In places we're taking a stricter reading of the BagIt spec in a way that I think is mostly fine – e.g. the BagIt spec says you SHOULD use the same algorithms for payload and tag manifests, but doesn't actually require it. This code does require they're the same.

(The BagIt spec also lists tag manifests as an optional element, but we require them.)

### DIFF
--- a/bag_verifier/src/test/scala/weco/storage_service/bag_verifier/services/BagVerifierTestCases.scala
+++ b/bag_verifier/src/test/scala/weco/storage_service/bag_verifier/services/BagVerifierTestCases.scala
@@ -351,7 +351,7 @@ trait BagVerifierTestCases[Verifier <: BagVerifier[
     }
   }
 
-  it("fails a bag if the file manifest does not exist") {
+  it("fails a bag if there are no payload manifests") {
     val badBuilder = new BagBuilderImpl {
       override protected def createPayloadManifest(
         entries: Seq[PayloadEntry]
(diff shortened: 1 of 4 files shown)


bag_verifier/src/test/scala/weco/storage_service/bag_verifier/services/BagVerifierTestCases.scala
```diff
@@ -726,34 +730,6 @@ trait BagVerifierTestCases[Verifier <: BagVerifier[
           )
       }
     }
-
-    it("passes a bag that includes an extra manifest/tag manifest") {
```
kenoir, line 730: Why did we have this test in the first place? Seems an unusual thing to worry about.

alexwlchan: Somewhere in the bag verifier there's a line that says _"look for any files in the bag that are (1) listed in one of the manifests or (2) the SHA-256 manifest"_.

Goobi uploads both SHA-256 and SHA-512 tag manifests, but until this BagReader patch we had to explicitly ignore for the extra tag manifest. We had to test the storage service wouldn't complain about a tag manifest it wasn't reading. I'm going to rework this a bit more in the verifier patch – it was easier to remove this test than get it working.

common/src/test/scala/weco/storage_service/bagit/services/BagReaderTestCases.scala
```diff
@@ -168,6 +341,170 @@ trait BagReaderTestCases[
...
+        bagInfo = bagInfo
+      )
+
+      storeBagContents(bagContents)
+
+      val err = withBagReader {
+        _.get(bagRoot).left.value
+      }
+
+      err.msg shouldBe "Tag manifests are inconsistent: each tag manifest should list the same set of tag files"
```
kenoir, line 399: It might be nice to give the consumer a clue about where to start looking i.e. highlight the first few things that are inconsistent without swamping the error response if everything is wrong. Not essential though.

kenoir (approved):

common/src/test/scala/weco/storage_service/bagit/services/BagReaderTestCases.scala
alexwlchan: Yeah, it's a good point. That gets complex enough that it should probably be separate, hmm.

PR #914, opened 2021-07-26
Use multiple checksums in the BagMatcher

alexwlchan:
Another tiny step towards #900.

Currently the `MatchedLocation` case class only holds a single checksum; now it holds every checksum in the bag and has a back-compatibility shim to return just one checksum. In the next patch I'll update the bag verifier to read and use this multi-checksum, but I'm still keeping these patches small so they can be easily reviewed.

### DIFF
--- a/bag_verifier/src/main/scala/weco/storage_service/bag_verifier/fixity/bag/BagExpectedFixity.scala
+++ b/bag_verifier/src/main/scala/weco/storage_service/bag_verifier/fixity/bag/BagExpectedFixity.scala
@@ -12,7 +12,6 @@ import weco.storage_service.bag_verifier.storage.{Locatable, Resolvable}
 import weco.storage_service.bag_verifier.storage.bag.BagLocatable
 import weco.storage_service.bagit.models._
 import weco.storage_service.bagit.services.BagMatcher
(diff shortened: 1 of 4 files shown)


bag_verifier/src/main/scala/weco/storage_service/bag_verifier/fixity/bag/BagExpectedFixity.scala
```diff
@@ -55,29 +54,25 @@ class BagExpectedFixity[BagLocation <: Location, BagPrefix <: Prefix[
     matched: MatchedLocation
   ): Either[Throwable, ExpectedFileFixity] =
     matched match {
-      case MatchedLocation(
-          bagPath: BagPath,
-          checksum: Checksum,
-          Some(fetchEntry)
-          ) =>
+      case loc @ MatchedLocation(bagPath, _, _, Some(fetchEntry)) =>
```
alexwlchan, line 57: In particular, here I’m going to use the multiChecksum from this unpacking and store that in the *FileFixity case classes, and then in the FixityChecker I'll be comparing that to the multiChecksum that we read.

kenoir (approved):

PR #923, opened 2021-08-02
Use the new per-account SQS DLQ alarm topic

alexwlchan:
This uses the new SQS DLQ alerting infra from https://github.com/wellcomecollection/platform-infrastructure/pull/221

If/when another message ends up on a DLQ in the storage service, we should now get alerts about it in Slack.


alexwlchan:
For https://github.com/wellcomecollection/platform/issues/5245

PR #937, opened 2021-09-07
Make the storage-service deployments more like the other services

alexwlchan:
* We get two new pipelines: deploy-stage and deploy-prod
* The stage pipeline is automatically triggered after a successful merge to main. It deploys services and runs the test bag, and waits for the test bag to complete (previously it would fire, but we'd only find out if it failed the following day).
* The prod pipeline is automatically triggered after a successful stage deploy, but waits for human intervention before deploying. If a human pushes the button, it deploys services and runs the test bag, waiting for the test bag to succeed.


.buildkite/scripts/send_and_wait_for_test_bag.py
```diff
@@ -0,0 +1,75 @@
...
+
+    return RequestsOAuthStorageServiceClient(
+        api_url=api_url,
+        client_id=client_id,
+        client_secret=client_secret,
+        token_url="https://auth.wellcomecollection.org/oauth2/token",
+    )
+
+
+if __name__ == "__main__":
```
kenoir, line 36: I think this will keep a CI instance awake and running for a long(ish) period. I don't think that is necessarily a problem, but would be good to quantify.

alexwlchan: Yeah, it's not dissimilar to what we do waiting for weco-deploy deployments.

I think we should look at running agents on differently sized instances, a la https://buildkite.com/docs/agent/v3/elastic-ci-aws#multiple-instances-of-the-stack

kenoir (approved):
This looks good to me, but we may want to consider having generalised deploy steps as with the Buildkite experience pipeline where we hand env params to a single "deploy" pipeline.

The advantage of having separate pipelines from my perspective is that you can visualise easily which env has failed in the Buildkite UI, but we may prefer the less-code, failure visibility in Slack model of the experience pipelines in general.

@jamieparkinson might have an opinion on this.

Don't think any of that blocks merging though.

jamieparkinson:
I am pro-parametrised-pipelines, but only when visibility is good enough in other ways - eg I'm still interested in using GitHub Deployments for everything. In the mean time this lgtm 👍

PR #958, opened 2021-11-03
Don't trigger infinite prod deploys

alexwlchan:
This step runs as part of the "deploy prod" pipeline… and triggers another "deploy prod" pipeline.

This is obviously wrong.


jamesgorrie (approved):
Interesting.

alexwlchan:
I think it's a copy/pasta from another pipeline.

PR #990, opened 2022-03-24
Allow Digirati to subscribe to notifications from the staging service

alexwlchan:
~This change is already applied; this PR is just for visibility.~

Okay it wasn't, but now it is.

I've also added a few notes on how the storage service Terraform is organised, and tried to make the cross-account SNS subscriptions a bit more obvious.


alexwlchan:
I'm going to merge this so I can apply it at the same time as #994.

Issue #993, opened 2022-04-11
Fix the issues around long-running bag replications


alexwlchan:
This is what I suspect to be the root cause of https://github.com/wellcomecollection/platform/issues/5484. I'm going to write it out with quite a lot of context, because (1) I don't know how quickly we'll implement this and (2) it's helpful for somebody else to understand it.

# Context

## What is the bag replicator?

The bag replicator is the service that copies a bag (a collection of files) from one storage location to another. It runs three times when storing a normal bag:

* Copying from working storage to our warm S3 storage
* Copying from working storage to our cold S3 storage
* Copying from working storage to our cold Azure storage

It calls ListObjects on the working storage, then does one-by-one copies of objects from the source to the destination.

## How the bag replicator tries to do write-once operation

We want the bag replicator to write exactly once to the permanent storage location – so we can replicate an object, verify it, then be certain that it hasn't changed after it was verified.

We don't currently enable the S3 features that allow you to do that (Legal Holds, Object Lock, governance mode, etc.). We do enable legal holds in Azure.

So instead, the bag replicator checks for an existing object itself, and only writes the object if there isn't already an object at the destination. If there's already an object, it confirms it's the same as what it's about to replicate.

```mermaid
flowchart TD
    F[get object to replicate] --> DE{does the <br/>destination have<br/> this object?}
    DE-- yes -->DEY{are the source <br/>and destination <br/>objects the same?}
    DE-- no -->DEN["✅ replicate the object"]
    DEY-- yes -->DEYY["✅ nothing to do"]
    DEY-- no -->DEYN["⚠️ something has gone wrong"]
```

To prevent two processes doing this check simultaneously, the bag replicator will lock around the destination prefix – if it's got the lock, it should be the only thing working on this prefix.

This uses our Dynamo locking code, which has locks that expire after a certain period. (This is so applications that crash without releasing their locks don't gum up a process in perpetuity.)

## S3 consistency and lazy checking

This approach had a bad interaction with the way S3 consistency used to work.

S3 used to be eventually consistent. If you called GetObject on a non-existent object, then PutObject, then GetObject again, you might get a 404 on the second call. This could cause verification failures, like so:

```mermaid
sequenceDiagram
  participant replicator
  participant verifier
  participant S3

  replicator->>S3: GetObject(f)
  S3->>replicator: 404 Not Found

  replicator->>S3: CopyObject(f, destination)
  S3->>replicator: OK, done!

  replicator->>verifier: you can verify f
[...]

PR #994, opened 2022-04-14
Reduce the Azure replicator concurrency to avoid intermittent issues

alexwlchan:
For https://github.com/wellcomecollection/storage-service/issues/993

This reduces the max tasks in the Azure replicator in prod from 10 to 8, plus makes sure we actually configure replicator locks properly (i.e. not the default of 3 minutes). We'll probably want to back out these changes again later, but I hope they'll stave off any more intermittent failures.


alexwlchan:
This is applied in prod.

PR #1001, opened 2022-05-30
Use a Localstack container for S3 instead of zenko/cloudserver

alexwlchan:
For https://github.com/wellcomecollection/platform/issues/5547

Note: I don't plan to merge this as-is; I'm using it to prove these tests can pass with localstack. Our localstack container runs on port 4566; our existing S3 container uses port 33333. I'd like to collapse it all into 4566, but that's defined in the S3Fixtures trait – I'll have to patch scala-libs to use the new port, then use the new version of the library here, then remove the overrides. This will also allow me to consolidate containers in some tests; if we have SQS+S3 in the same test, we can do them both in a single localstack container.

### DIFF
--- a/bag_register/docker-compose.yml
+++ b/bag_register/docker-compose.yml
@@ -5,8 +5,8 @@ localstack:
   ports:
     - "4566:4566"
 s3:
-  image: "760097843905.dkr.ecr.eu-west-1.amazonaws.com/zenko/cloudserver:8.1.8"
+  image: "760097843905.dkr.ecr.eu-west-1.amazonaws.com/localstack/localstack"
   environment:
-    - "S3BACKEND=mem"
+    - SERVICES=s3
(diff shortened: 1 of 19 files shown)


alexwlchan:
Some of the tests failed for things that look like they're fixed in newer versions of the Localstack image (e.g. expiry for pre-signed S3 URLs); I've pushed a new image to see if that makes the tests happier.

alexwlchan:
And I immediately broke all the tests, damnit: https://github.com/localstack/localstack/issues/5894

jamieparkinson (approved):

alexwlchan:
It looks like we're hitting https://github.com/localstack/localstack/issues/5894, so I'm going to try rolling back to a slightly older Localstack image until I can sort it in scala-libs.

PR #1023, opened 2022-09-09
Fix reindexing scripts

jamieparkinson:
This is not my finest work, but it has solved our immediate problems. It was mostly moving existing code around and tweaking it until it works.


indexer/reindexing/chunked_diff.py
```diff
@@ -0,0 +1,18 @@
...
+    def _chunks(big_list, chunk_length):
+        for i in range(0, len(big_list), chunk_length):
+            yield big_list[i : i + chunk_length]
+
+    chunk_length = 500
+    chunk_count = math.ceil(len(all_entries) / chunk_length)
+
+    diff_list = []
+    for chunk in tqdm(_chunks(all_entries, chunk_length), total=chunk_count):
+        diff_list.append(diff_for_chunk(chunk))
```
paul-butcher, line 16: I think this could be a generator, 
```suggestion
    diff_list = (diff_for_chunk(chunk) for chunk in tqdm(_chunks(all_entries, chunk_length), total=chunk_count))
```
it's only iterated once just below.

If not, a list comprehension could work.

jamieparkinson: Makes sense to me, absolutely no recollection of writing this!

paul-butcher (approved):

jamieparkinson:
Glad this PR has had a whole year to mature now

Issue #1061, opened 2023-05-18
Remove a handful of images from the storage service


alexwlchan:

alexwlchan:
I'm going to leave this open as I still need to finish the documentation for it.

alexwlchan:
There's an open PR to improve the documentation; once that's reviewed and merged, this can be closed.

PR #1063, opened 2023-05-25
Explain how to delete bags

alexwlchan:
Closes https://github.com/wellcomecollection/storage-service/issues/1061


scripts/helpers/storage_service.py
```diff
@@ -33,3 +38,35 @@ def lookup_ingest(ingest_id):
...
+
+    matching_manifests = list(
+        list_s3_prefix(
+            s3,
+            bucket=manifests_bucket,
+            prefix=f"{space}/{external_identifier}/{version.replace('v', '')}/",
+        )
+    )
+
+    if len(matching_manifests) >= 1:
```
paul-butcher, line 61: I think empty lists are falsy.

scripts/helpers/azure.py
```diff
@@ -209,5 +209,16 @@ def delete_azure_prefix(*, account, container, prefix):
...
+    # of doing it, but:
+    #
+    #   1.  I'm not super familiar with the Azure CLI; when I tried the
+    #       command for deleting multiple objects at once, it just hung
+    #       forever.
+    #   2.  If you try to be clever and run multiple instances of the CLI
+    #       at once, you hit rate limits pretty quickly.
+    #
+    # Given deletions are extremely rare, I'm okay with this being a
+    # somewhat slow process.
```
paul-butcher, line 222: Great commentary!

scripts/helpers/storage_service.py
alexwlchan: Good catch!

alexwlchan: Ah yeah, I remember, this was originally `if len(matching_manifests) == 1` and it would throw if it found more than one, but it's possible for us to create two manifests if the bag register (which creates the manifest) runs twice – but all the manifests should always have the same ingest ID, so picking any of them is fine.

Issue #1037, opened 2022-12-05
Support periods in external identifiers, e.g. PP/GRF/A.41


alexwlchan:
As part of the born-digital work, Ashley has found some CALM identifiers that use periods, e.g. `PP/GRF/A.41`

The storage service won't store these, because it doesn't allow periods in external identifiers. If you try, it fails with an error in the ingests API:

> 15:27:19.897 [main-actor-system-akka.actor.default-dispatcher-2921] ERROR weco.http.WellcomeHttpApp - Sending HTTP 500: java.lang.IllegalArgumentException: requirement failed: External identifier must match regex: ^[a-zA-Z0-9_\-/ ]+$

We want it to mirror catalogue identifiers as much as possible (see [How identifiers work in the storage service](https://docs.wellcomecollection.org/storage-service/reference-design-decisison/identifiers)), so we should support this.

This means:

* [x] Updating the ExternalIdentifier model to support periods (plus a test)
* [x] Improving the error message on ExternalIdentifier to show the identifier in question
* [x] Fixing the ingests API so it returns an HTTP 400 for bad identifiers, not HTTP 500
* [x] Adding a test that the bags API can retrieve a bag with a period in the identifier

This would be a good first patch for somebody – it's not too complicated, it's not time-sensitive, and it touches a bunch of bits of the storage service. I'd like to pair on this with somebody else.

alexwlchan:
An additional thought: we should support periods _in the middle_ of identifiers, but at the beginning or end is liable to cause issues e.g. with hidden files on macOS or Azure keys.

paul-butcher:
Are there any other characters that should have positional restrictions?  Not necessarily for this change, but I'm concerned about the way spaces are included and the possibility that `       ` and `   ` are two different acceptable identifiers, as are `PP GRF X` and `PP  GRF X`

alexwlchan:
> Are there any other characters that should have positional restrictions?

Good question! Spaces probably should have positional restrictions, and I think slashes already do? (If not they should!)

alexwlchan:
Looking at the regex (https://github.com/wellcomecollection/storage-service/blob/cc76ddb621474e89d6cf4dec489c2120466b0cc7/common/src/main/scala/weco/storage_service/bagit/models/ExternalIdentifier.scala#L81), I think a safer rule might be: **external identifiers must start and end with an alphanumeric character**.

PR #1098, opened 2023-09-08
New id constraints

paul-butcher:
See: https://github.com/wellcomecollection/storage-service/issues/1037

This adds the ability to use identifiers containing dots, and also tightens the rules around spaces and start/end characters.

### DIFF
--- a/ingests/ingests_api/src/main/scala/weco/storage_service/ingests_api/responses/CreateIngest.scala
+++ b/ingests/ingests_api/src/main/scala/weco/storage_service/ingests_api/responses/CreateIngest.scala
@@ -4,8 +4,15 @@ import akka.http.scaladsl.model.StatusCodes
 import akka.http.scaladsl.model.headers.Location
 import akka.http.scaladsl.server.Route
 import grizzled.slf4j.Logging
-import weco.storage_service.ingests.models.{AmazonS3StorageProvider, Ingest, StorageProvider}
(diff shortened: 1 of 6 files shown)


ingests/ingests_api/src/main/scala/weco/storage_service/ingests_api/responses/CreateIngest.scala
```diff
@@ -65,11 +68,37 @@ trait CreateIngest[UnpackerDestination] extends FutureDirectives with Logging {
...
+    Try {
+      requestDisplayIngest.toIngest
+    } match {
+      case Failure(exception) =>
+        exception match {
+          case invalidArg: IllegalArgumentException =>
+            convertToBadRequestResponse(invalidArg)
+          case otherException => throw otherException
+        }
+      case Success(ingest) => triggerIngestStarter(ingest)
```
alexwlchan, line 97: I wonder if we could collapse this into a single `case` statement:
```suggestion
      case Failure(invalidArg: IllegalArgumentException) =>
        convertToBadRequestResponse(invalidArg)
      case Failure(exception) => throw exception
      case Success(ingest) => triggerIngestStarter(ingest)
```

alexwlchan (approved):
Overall this is really nice; I like all the ways you've made it even stricter and the new test descriptions are great. 💪 

One small suggestion about tidying up some nested switch statements, but otherwise looks great.

PR #1122, opened 2024-04-09
Use Common Runtime Libraries in S3TransferManager

kenoir:
## What does this change?

This change attempts to make use of the [Common Runtime Libraries](https://docs.aws.amazon.com/sdkref/latest/guide/common-runtime.html) for the AWS SDK intended to improve performance. 

We [include them as a dependency](https://github.com/wellcomecollection/storage-service/blob/main/project/Dependencies.scala#L110) but do not use them in the `S3TransferManager` where we do large transfers. The [documentation describes how to do this](https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/transfer/s3/S3TransferManager.html).

The intention is to improve the performance of the replicators which are currently showing issues transferring large files to S3.

## How to test

- [x] Do the tests pass?
- [ ] When deployed do bags with large files get replicated succesfully?

Improved performance, less failures.

This will impact replication performance, hopefully only positively but there are could be negative impacts.

### DIFF
--- a/bag_replicator/docker-compose.yml
+++ b/bag_replicator/docker-compose.yml
@@ -1,17 +1,19 @@
-localstack:
-  image: "760097843905.dkr.ecr.eu-west-1.amazonaws.com/localstack/localstack"
-  environment:
-    - SERVICES=sqs
-  ports:
-    - "4566:4566"
-s3:
-  image: "760097843905.dkr.ecr.eu-west-1.amazonaws.com/zenko/cloudserver:8.1.8"
-  environment:
-    - "S3BACKEND=mem"
-  ports:
-    - "33333:8000"
-azurite:
-  image: "mcr.microsoft.com/azure-storage/azurite"
-  ports:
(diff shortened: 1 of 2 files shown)


bag_replicator/docker-compose.yml
```diff
@@ -1,17 +1,19 @@
...
-  environment:
-    - "S3BACKEND=mem"
-  ports:
-    - "33333:8000"
-azurite:
-  image: "mcr.microsoft.com/azure-storage/azurite"
-  ports:
-    - "10000:10000"
-  command: ["azurite", "--blobHost", "0.0.0.0"]
+version: "3.3"
```
kenoir, line 1: Needed to update this to run with my version of docker, think that will be the case for anyone wanting to run this.

kenoir:
Closing as this doesn't change the implementation (the default behaviour is this hidden in the s3 transfer manager builder.

PR #1136, opened 2024-08-15
Update links to ingest inspector from daily reporter lambda

kenoir:
## What does this change?

This updates the daily reporter lambda that provides summaries of storage service activities in Slack to use the correct URL pattern for the new ingest inspector. In addition this PR adds a GitHub action workflow to deploy the daily reporter lambda as this previously was done manually.

## How to test

- [ ] Manually trigger the new workflow and invoke the updated lambda from the AWS console, does the updated link appear?

The daily reporter lambda provides working links to the ingest inspector enabling quicker debugging and future updates to the daily-reporter are deployed automatically on merge to main.

This change should not impact any critical storage service processes and can be safely reverted if it does not behave as expected.

### DIFF
--- a/.github/workflows/deploy-daily-reporter.yml
+++ b/.github/workflows/deploy-daily-reporter.yml
@@ -1,0 +1,26 @@
+name: Deploy Daily Reporter
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - main
+
+permissions:
+  id-token: write
+
+jobs:
+  deploy_ingest_inspector:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: aws-actions/configure-aws-credentials@v4
+        with:
+          aws-region: eu-west-1
+          role-to-assume: ${{ secrets.DEPLOY_INGEST_INSPECTOR_GHA_ROLE_ARN }}
+      - name: Deploy Daily Reporter
+        run: |
+          pip3 install -r ./builds/requirements.txt
+          python3 ./builds/publish_lambda_zip.py monitoring/daily_reporter \
+              --bucket=wellcomecollection-storage-infra \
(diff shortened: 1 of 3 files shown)


.github/workflows/deploy-daily-reporter.yml
```diff
@@ -0,0 +1,26 @@
...
+
+jobs:
+  deploy_ingest_inspector:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: aws-actions/configure-aws-credentials@v4
+        with:
+          aws-region: eu-west-1
+          role-to-assume: ${{ secrets.DEPLOY_INGEST_INSPECTOR_GHA_ROLE_ARN }}
```
kenoir, line 19: This role I think has sufficient permissions to perform this deployment and is arguably the correct role to use in any case given the relationship between the daily reporter and the ingest inspector.

monitoring/daily_reporter/src/slack.py
```diff
@@ -29,7 +29,7 @@ def _get_slack_message(label, ingests):
         if ingests.get(status, []):
             result += "\n" + status.title() + ":"
             for i in ingests[status][:15]:
-                result += f"\n- <https://ingest-inspector.wellcomecollection.org/ingests/{i['id']}|`{i['id']}`> – {i['space']}/{i['externalIdentifier']}"
+                result += f"\n- <https://ingest-inspector.wellcomecollection.org/?ingestId={i['id']}|`{i['id']}`> – {i['space']}/{i['externalIdentifier']}"
```
kenoir, line 32: The URL that the daily reporter currently uses is that of the Glitch ingest inspector. Rather than the modified one here (which also does not resolve). I think the deployment for the daily reporter of the last mosst recent change was missed due to the lack of automation.

paul-butcher (approved):

PR #1156, opened 2025-04-22
delete_azure_blobs_concurrently

agnesgaroux:
## What does this change?

Part of https://github.com/wellcomecollection/platform/issues/5942
The Azure blob deletion can take a long time.
This replaces the CLI with the python SDK, with 30 threads running in parallel 
The plan was to start with a number of workers that was probably too high, and decrease it if it errored but it's fine with 30 💪

## How to test

Ran in staging for 2 bags of similar contents:
Screenshot 2025-04-22 at 14 32 30
Screenshot 2025-04-22 at 14 35 12
The first one was deleted in about a minute (with parallelism)
The other is still hanging as I'm opening this PR, it's been almost 20 minutes. UPDATE it ended taking about 22-23 minutes (without parallelism)

Faster Azure blobs deletions, less time staring at the terminal


kenoir (approved):
My understanding is that mapping over the thread pool executor will halt if there are any exceptions, which is fine (as long as there isn't silent failure it's good).

agnesgaroux:
> My understanding is that mapping over the thread pool executor will halt if there are any exceptions, which is fine (as long as there isn't silent failure it's good).

I think any exception would be caught in the try/except block that wraps `container_client.delete_blob(blob_name)`

PR #1177, opened 2026-05-06
scripts: migrate to uv for dependency management (RFC 071)

kenoir:
## What does this change?

Migrates the `scripts/` folder to [uv](https://github.com/astral-sh/uv) for dependency and environment management, in line with [RFC 071: Python Building and Deployment](https://github.com/wellcomecollection/docs/tree/main/rfcs/071-python_builds).

**Changes:**

- Adds `scripts/pyproject.toml` — declares runtime dependencies (boto3, click, daiquiri, elasticsearch, humanize, tenacity, azure-storage-blob) and a `[dependency-groups.test]` group (pytest, pytest-cov, coverage, moto). Declares `wellcome-storage-service` as an editable local dependency from `../python_client`, so the in-repo client is always used without a separate install step.
- Adds `scripts/uv.lock` — reproducible lockfile so all contributors and CI get identical environments.
- Adds `scripts/README.md` — documents how to set up the environment and run scripts and tests using uv.
- Removes `scripts/requirements.in`, `scripts/requirements.txt`, `scripts/test_requirements.in`, `scripts/test_requirements.txt` — superseded by pyproject.toml and uv.lock.
- Adds `.metals/` to `.gitignore`.

## How to test

From `scripts/`:

```
# Install runtime deps
uv sync

# Run a script (e.g.)
uv run ss_get_ingest.py <INGEST_ID>

# Install test deps and run tests
uv sync --group test
uv run --group test pytest -q
```

All 34 existing tests pass.

- `uv sync` completes without error in a clean checkout.
- `uv run --group test pytest -q` reports all tests green.
- No `pip install` or `pip-compile` steps are needed to work with this folder.

- The `../python_client` editable source path assumes the `scripts/` folder is always used from within the `storage-service` repository clone. This was already implicit before — the package was installed manually or via buildkite pip invocations. The path is consistent with the RFC's recommendation to treat co-located shared code as an installable library.
- `uv.lock` pins exact versions; dependabot or a periodic `uv lock --upgrade` will be needed to keep dependencies current.


scripts/pyproject.toml
```diff
@@ -0,0 +1,30 @@
+[project]
+name = "scripts"
+version = "0.1.0"
+description = "Add your description here"
```
agnesgaroux, line 4: "Add your description here" 😅

agnesgaroux (approved):

PR #1194, opened 2026-08-13
stack_staging: upgrade AWS provider to 5.100, matching stack_prod

kenoir:
## What does this change?

stack_staging keeps its own lock file, so the provider upgrade in #1193 only cleared prod. Staging was still on hashicorp/aws 5.35.0 from January 2024, where a full plan proposes replacing around 150 inline IAM role policies: they were created in 2020 with an auto-recorded `name_prefix` that the config never sets, and old 5.x versions read that as drift. That makes any whole-stack apply unsafe. This is the follow-up flagged on #1193.

The lock moves to 5.100.0 with hashes for darwin_arm64, darwin_amd64 and linux_amd64, and the root gains a `~> 5.0` `required_providers` pin, since nothing else constrained the provider here.

## How to test

With terraform 1.6.5 and provider 5.100.0, a full `terraform plan` in `stack_staging` shows the replicator memory change from #1193 that has not been applied to staging yet (three task definition replacements, three service updates), plus two in-place lifecycle normalisations where the newer provider drops an empty `filter` block. No IAM policy churn.

Staging can be applied as a whole stack again rather than resource by resource, and its plan matches prod's.

This has not been applied to staging yet, and applying it will also roll out the replicator memory increase from #1193. That is a rolling replacement of three services; in-flight messages return to the queue and the replicators are idempotent. The provider jump is the larger risk, which is why the full plan was read rather than a targeted one.

### DIFF
--- a/terraform/stack_staging/.terraform.lock.hcl
+++ b/terraform/stack_staging/.terraform.lock.hcl
@@ -2,23 +2,26 @@
 # Manual edits may be lost in future updates.
 
 provider "registry.terraform.io/hashicorp/aws" {
-  version = "5.35.0"
+  version     = "5.100.0"
+  constraints = "~> 5.0"
   hashes = [
-    "h1:MKNFmhsOIirK7Qzr6TWkVaBcVGN81lCU0BPiaPOeQ8s=",
-    "zh:3a2a6f40db82d30ea8c5e3e251ca5e16b08e520570336e7e342be823df67e945",
-    "zh:420a23b69b412438a15b8b2e2c9aac2cf2e4976f990f117e4bf8f630692d3949",
-    "zh:4d8b887f6a71b38cff77ad14af9279528433e279eed702d96b81ea48e16e779c",
-    "zh:4edd41f8e1c7d29931608a7b01a7ae3d89d6f95ef5502cf8200f228a27917c40",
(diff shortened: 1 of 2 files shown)


agnesgaroux:
Does it have to be `terraform 1.6.5`?

agnesgaroux:
Getting 
`Plan: 3 to add, 5 to change, 3 to destroy.`
-> 3 replace (ie. 3 to destroy and 3 to add) and 3 add for the memory change
-> unpacked bags and large response cache buckets lifecycle configuration, dropping `filter`

agnesgaroux (approved):
Bearing in mind I've run with `Terraform v1.10.1` 
(which requires the assume_role wrapper, I have it on my local branch if you want me to push it on top of this)

kenoir:
Good point, thanks. I've pushed the assume_role conversion for stack_staging (backend block and all five remote state data sources) in 64dfb36f2 rather than have you rebase your local branch. Verified terraform init and validate on both 1.6.5 (which last wrote the state) and 1.14.4, and the plan is unchanged. The other stacks still use the legacy syntax; I'll sweep those in a follow-up so this PR stays scoped to staging. One thing to coordinate: the first apply from a newer binary stamps the state with its version and older binaries then refuse to write, so it's worth us agreeing a single team version when we pick this up.

